PR #13, opened 2026-02-19
feat(client): add RolloutClient with S3 HEAD polling

luyuzhe111:
*Description of changes:*

- Add RolloutFuture with per-future exponential backoff
- Add BatchResult iterator for streaming batch results
- Add BatchItem wrapper for success/error distinction
- TPS rate limiting and session concurrency control

With this PR, UX will be streamlined to the following 

```python

from agentcore_rl_toolkit import RolloutClient

client = RolloutClient(...)

# client acknowledges the payload is received, returning the s3 bucket and result key
future = client.invoke(payload)

# under the hood, future.result() repeatedly poll S3 with HEAD to check its existence
result = future.result()
```

SQS will be optional and removed later so users don't have to worry about it. 

### DIFF
--- a/src/agentcore_rl_toolkit/__init__.py
+++ b/src/agentcore_rl_toolkit/__init__.py
@@ -1,6 +1,16 @@
 from .app import AgentCoreRLApp
+from .client import BatchItem, BatchResult, RolloutClient, RolloutFuture
 from .frameworks.strands import StrandsAgentCoreRLApp
 from .frameworks.strands.rollout_collector import RolloutCollector as StrandsRolloutCollector
 from .reward_function import RewardFunction
 
-__all__ = ["AgentCoreRLApp", "StrandsAgentCoreRLApp", "StrandsRolloutCollector", "RewardFunction"]
+__all__ = [
+    "AgentCoreRLApp",
+    "StrandsAgentCoreRLApp",
+    "StrandsRolloutCollector",
+    "RewardFunction",
+    "RolloutClient",
+    "RolloutFuture",
+    "BatchResult",
+    "BatchItem",
+]
--- a/src/agentcore_rl_toolkit/client.py
+++ b/src/agentcore_rl_toolkit/client.py
@@ -121,15 +121,32 @@ class RolloutClient:
         concurrent access from multiple threads.
     """
 
+    @staticmethod
+    def _parse_region_from_arn(arn: str) -> str:
+        """Extract AWS region from an ARN.
+
+        ARN format: arn:partition:service:region:account-id:resource-type/resource-id
+
+        Args:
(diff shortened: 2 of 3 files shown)


src/agentcore_rl_toolkit/__init__.py
lliquid, line 1: @luyuzhe111  can we add code regarding UX change as well?

src/agentcore_rl_toolkit/client.py
```diff
@@ -0,0 +1,347 @@
...
+            # Use the per-future backoff interval
+            time.sleep(self._poll_interval)
+
+
+class RolloutClient:
+    """Client for invoking agents and collecting rollouts with full lifecycle management.
+
+    Note:
+        This client is NOT thread-safe. Use separate client instances for
+        concurrent access from multiple threads.
```
lliquid, line 121: Under what condition will there be multi-thread access to the same rollout client?

src/agentcore_rl_toolkit/client.py
```diff
@@ -0,0 +1,347 @@
...
+
+    def __init__(
+        self,
+        agent_runtime_arn: str,
+        s3_bucket: str,
+        exp_id: str,
+        region: str = None,
+        max_retry_attempts: int = 5,
+        # ACR constraints
+        max_concurrent_sessions: int = 100,
```
lliquid: i am wondering if those parameters can be retrieved from agentcore instead of being manually set.

src/agentcore_rl_toolkit/__init__.py
luyuzhe111: @lliquid did you mean "comment" actually? was planning to add documentations in the PR that removes SQS. what do you think?

src/agentcore_rl_toolkit/client.py
luyuzhe111: basically if the user runs client.run_batch() simultaneously, the rate and concurrency limiting will be unreliable since `client._last_invoke_time` and `client.max_concurrent_sessions` won't balance between multiple batches.

src/agentcore_rl_toolkit/client.py
luyuzhe111: ah that's a good point. for `region`, it can & should be extracted from the agent arn. I have fixed in the new commit. for `max_concurrent_sessions`, since it's actually highly dependent on the user's quota (number of agentcore runtime sessions, model API quotas from those agent sessions, etc), I've changed it to a required parameter. However, it's no longer required in the constructor, but rather in `client.run_batch()` method, so users only specify it when they need it. Does these changes make sense?

lyzustc (approved):

PR #14, opened 2026-02-20
feat(examples): add code migration agent

luyuzhe111:
*Description of changes:* This PR adds a code migration agent example. This example is under active development alongside the `agentcore-rl-toolkit` library.


src/agentcore_rl_toolkit/frameworks/strands/vllm_model.py
```diff
@@ -0,0 +1,107 @@
+"""vLLM model provider with token ID collection for RL training."""
```
lyzustc, line 1: vLLM model may be used by many examples, how about move it to src/agentcore_rl_toolkit/frameworks/strands/

lyzustc (commented):
Suggest to move vllm model class to src/agentcore_rl_toolkit/frameworks/strands so other examples could use it.

src/agentcore_rl_toolkit/frameworks/strands/vllm_model.py
luyuzhe111: @lyzustc just moved to `src/agentcore_rl_toolkit/frameworks/strands/`! originally planning it as a tmp solution but I guess it wouldn't hurt. thanks!

lyzustc (approved):

PR #21, opened 2026-02-25
docs(examples): migrate docs + examples to AgentCoreRLApp with vLLMModel

luyuzhe111:
*Description of changes:*
- Replace StrandsAgentCoreRLApp/StrandsRolloutCollector with the framework-agnostic AgentCoreRLApp and vLLMModel for direct token collection. 
- Remove SQS dependency in favor of S3-only result delivery.
- Make S3 config optional in _rollout payload while requiring base_url and model_id. 
- Add rationale for per-invocation model/agent creation.

### DIFF
--- a/examples/strands_math_agent/rl_app.py
+++ b/examples/strands_math_agent/rl_app.py
@@ -1,55 +1,60 @@
-from dotenv import load_dotenv
 from reward import GSM8KReward
 from strands import Agent
 from strands_tools import calculator
 
-from agentcore_rl_toolkit import StrandsAgentCoreRLApp, StrandsRolloutCollector
+from agentcore_rl_toolkit import AgentCoreRLApp
+from agentcore_rl_toolkit.frameworks.strands.vllm_model import vLLMModel
 
-app = StrandsAgentCoreRLApp()
+app = AgentCoreRLApp()
 
-load_dotenv()
-
-model = app.create_openai_compatible_model()
-
-rollout_collector = StrandsRolloutCollector()
-agent = Agent(
-    model=model,
-    tools=[calculator],
-    system_prompt=(
-        "Your task is to solve the math problem. "
-        + "Use the calculator tool to compute all mathematical expressions. "
-        + 'Let\'s think step by step and output the final answer after "####".'
-    ),
-    hooks=[rollout_collector],
+system_prompt = (
+    "Your task is to solve the math problem. "
+    + "Use the calculator tool to compute all mathematical expressions. "
+    + 'Let\'s think step by step and output the final answer after "####".'
 )
+
 reward_fn = GSM8KReward()
 
 
 @app.rollout_entrypoint
-async def invoke_agent(payload, context):
+def invoke_agent(payload: dict):
     """
     Invoke the math agent with a payload using the rollout_entrypoint decorator.
 
     For RL training, the following fields are expected:
     - prompt: question from gsm8k
     - answer: ground truth (str)
+    - _rollout: rollout config with base_url and model_id
 
     The @rollout_entrypoint decorator automatically:
-    - Handles asyncio.create_task() for non-blocking execution
-    - Saves rollout data to S3 and notifies SQS when returned
-    - Handles errors and saves error rollouts for client notification
+    - Executes the function in the background for non-blocking processing
+    - Saves rollout data to S3 with a predictable key
+    - Handles errors and saves error rollouts for client awareness
     - Works with both sync and async functions
     """
+    base_url = payload["_rollout"]["base_url"]
+    model_id = payload["_rollout"]["model_id"]
+
+    model = vLLMModel(
+        client_args={"api_key": "abc", "base_url": base_url},
+        model_id=model_id,
+    )
+
+    agent = Agent(
+        model=model,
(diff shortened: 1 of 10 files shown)


examples/strands_migration_agent/dev_app.py
```diff
@@ -39,8 +39,8 @@
 
 @app.rollout_entrypoint
 def invoke_agent(payload: dict):
-    base_url = payload.get("_rollout", {}).get("base_url")
-    model_id = payload.get("_rollout", {}).get("model_id")
+    base_url = payload["_rollout"]["base_url"]
```
lyzustc, line 42: Here you want agent app to directly report error if payload does not have them?

src/agentcore_rl_toolkit/__init__.py
```diff
@@ -1,13 +1,9 @@
 from .app import AgentCoreRLApp
 from .client import BatchItem, BatchResult, RolloutClient, RolloutFuture
-from .frameworks.strands import StrandsAgentCoreRLApp
-from .frameworks.strands.rollout_collector import RolloutCollector as StrandsRolloutCollector
 from .reward_function import RewardFunction
 
 __all__ = [
     "AgentCoreRLApp",
-    "StrandsAgentCoreRLApp",
```
lyzustc, line 9: We could remove their implementations if they are no longer needed.

examples/strands_math_agent/rl_app.py
```diff
@@ -1,55 +1,60 @@
...
     """
+    base_url = payload["_rollout"]["base_url"]
+    model_id = payload["_rollout"]["model_id"]
+
+    model = vLLMModel(
+        client_args={"api_key": "abc", "base_url": base_url},
+        model_id=model_id,
+    )
+
+    agent = Agent(
```
lliquid, line 43: @luyuzhe111  can the agent be defined outside of invoke? this feels more natural in terms of users mental model

lyzustc (approved):

examples/strands_math_agent/rl_app.py
```diff
@@ -1,55 +1,60 @@
...
-    response = await agent.invoke_async(user_input)
+    response = agent(user_input)
 
-    # Gather rollouts from the collector
-    rollout_data = rollout_collector.get_rollout_data()
+    # Collect token data (prompt IDs, response IDs, logprobs) from the model
+    rollout_data = model.get_token_data()
 
     # Compute rewards
     rewards = reward_fn(response_text=response.message["content"][0]["text"], ground_truth=answer)
```
lliquid, line 60: @luyuzhe111  i am wondering if we can have a defined data class for reward and also a consistent interface regarding what the reward function takes. in this way developer will have a good idea about how to write the reward function

examples/strands_migration_agent/dev_app.py
```diff
@@ -39,8 +39,8 @@
...
+    model_id = payload["_rollout"]["model_id"]
 
     model = vLLMModel(
         client_args={"api_key": "abc", "base_url": base_url}, model_id=model_id, params={"max_completion_tokens": 8192}
```
lliquid: params={"max_completion_tokens": 8192} and "api_key": "abc" looks like magic numbers/params

src/agentcore_rl_toolkit/__init__.py
luyuzhe111: yep will do in a cleanup PR!

examples/strands_migration_agent/dev_app.py
luyuzhe111: yea this will give clear errors. otherwise the error would be a bit magical (model not found on openai, something like that).

examples/strands_math_agent/rl_app.py
luyuzhe111: agreed. but defining globally has the following limitations:

- we can't pass in different sampling parameters for training vs validation, which is a common practice. Otherwise, we have to ask users to find a way to update params, a mechanism may or may not exist. 
- we will have to deploy the agent to agentcore in the training framework, since the inference server (and possibly model id, which can be an absolutely path to the checkpoint) can vary between runs. this means we will have to bring all kinds of configs, such as ECR image, vpc config, etc to the training framework. If we define the model in the invocation path, in contrast, we can deploy the agent beforehand and just provide agent arn to the training framework (ux).
- we can't adapt to different learning paradigms (e.g., system prompt optimization) if we define the agent globally. 

so given the trade-off, i feel like defining the agent in the invocation entrypoint is the best option we have. i also added explanations to README regarding this difference. let me know what you think!

examples/strands_math_agent/rl_app.py
luyuzhe111: this is tricky since reward function can be really diverse. but i think in the future we can provide some reward function template, for LLMJ for example, which seems to have more fixed patterns?

examples/strands_migration_agent/dev_app.py
luyuzhe111: good catch. fixed in a new commit. replaced "abc" with vllm convention "empty" for both examples. sampling params will be passed as a whole from the client side!

PR #28, opened 2026-03-04
feat(client): add async methods to RolloutClient and RolloutFuture

luyuzhe111:
*Description of changes:*

Implements invoke_async, run_batch_async, result_async, done_async, and __await__ enable non-blocking usage in asyncio event loops for RolloutClient. Uses asyncio.to_thread for boto3 calls.

Also extracts shared eval helpers into eval_utils.py and adds evaluate_async.py example demonstrating both batch and individual async patterns.

### DIFF
--- a/examples/strands_migration_agent/eval_utils.py
+++ b/examples/strands_migration_agent/eval_utils.py
@@ -1,0 +1,70 @@
+"""Shared utilities for evaluation scripts."""
+
+import json
+from pathlib import Path
+
+import boto3
+import tomllib
+
+
+def load_config(config_path: str = "config.toml") -> dict:
+    """Load configuration from TOML file if it exists."""
+    path = Path(__file__).parent / config_path
+    if path.exists():
+        with open(path, "rb") as f:
+            return tomllib.load(f)
+    return {}
+
+
+def get_s3_folder_uris(s3_uri: str) -> list[str]:
+    """
+    Get full S3 URIs for folders under a specific S3 path.
+
+    Args:
+        s3_uri: Can be full URI (s3://bucket/prefix/) or just bucket name
+    """
+    path = s3_uri.replace("s3://", "")
+    bucket_name = path.split("/")[0]
+    prefix = "/".join(path.split("/")[1:])
+
+    s3 = boto3.client("s3")
+
+    folder_uris = []
+    paginator = s3.get_paginator("list_objects_v2")
+
+    for page in paginator.paginate(Bucket=bucket_name, Prefix=prefix, Delimiter="/"):
+        for prefix_obj in page.get("CommonPrefixes", []):
+            folder_path = prefix_obj["Prefix"]
+            folder_uris.append(f"s3://{bucket_name}/{folder_path}")
+
+    return folder_uris
+
+
+def prepare_payload(folder_uri: str) -> dict:
+    """
+    Prepare a single payload for a repository folder.
+
+    Args:
+        folder_uri: S3 folder URI like "s3://migration-bench/15093015999__EJServer/"
+
+    Returns:
+        Payload dictionary (without _rollout config, RolloutClient adds that)
+    """
+    folder_uri = folder_uri.rstrip("/")
+    repo_name = folder_uri.split("/")[-1]
+
+    repo_uri = f"{folder_uri}/{repo_name}.tar.gz"
+    metadata_uri = f"{folder_uri}/metadata.json"
+
+    return {
+        "prompt": "Please help migrate this repo: {repo_path}. There are {num_tests} test cases in it.",
+        "repo_uri": repo_uri,
+        "metadata_uri": metadata_uri,
+        "require_maximal_migration": False,
+    }
+
+
+def append_result_to_file(result_path: Path, item_data: dict):
+    """Append a single result to the JSON file (one JSON object per line)."""
+    with open(result_path, "a") as f:
+        f.write(json.dumps(item_data) + "\n")
--- a/src/agentcore_rl_toolkit/client.py
+++ b/src/agentcore_rl_toolkit/client.py
@@ -1,5 +1,6 @@
 """Client for invoking agents and collecting rollouts via S3 polling."""
 
+import asyncio
 import json
 import logging
 import time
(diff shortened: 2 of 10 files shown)


lyzustc (approved):

examples/strands_migration_agent/eval_utils.py
```diff
@@ -0,0 +1,70 @@
+"""Shared utilities for evaluation scripts."""
```
lyzustc, line 1: @luyuzhe111 I think in the future, every agent example will need a batch evaluation script, so the commonly used evaluation tool functions could be put into package source code library.

luyuzhe111: hi @lyzustc, are you referring to functions in this file specifically? it's specific to this example tho. but yea if there are shared eval utilities in the future we can add them to the source package!

lyzustc: ok, that makes sense.

src/agentcore_rl_toolkit/client.py
```diff
@@ -132,6 +133,63 @@ def cancelled(self) -> bool:
...
+    async def done_async(self) -> bool:
+        """Check if result is ready (non-blocking, async). Updates backoff state.
+
+        Like ``done()`` but wraps the S3 HEAD call in ``asyncio.to_thread()``
+        so it doesn't block the event loop.
+        """
+        if self._done or self._cancelled:
+            return True
+        try:
+            await asyncio.to_thread(self.s3_client.head_object, Bucket=self.s3_bucket, Key=self.result_key)
```
hellodanylo, line 151: Does this create a new thread for each .done() call? It's highly undesirable to create more than a few dozen threads per Python process. If that's the case, I recommend either an async version of boto3, or create a ThreadPool to limit the count of threads used.

src/agentcore_rl_toolkit/client.py
```diff
@@ -132,6 +133,63 @@ def cancelled(self) -> bool:
...
+    async def _async_poll(self) -> dict:
+        """Poll until result is ready, using asyncio.sleep for backoff."""
+        if self._cancelled:
+            raise CancelledError("Future was cancelled")
+        if self._result is not None:
+            return self._result
+
+        while True:
+            if await self.done_async():
+                self._result = await asyncio.to_thread(self._fetch_result)
```
hellodanylo, line 175: Same question here as for the .done(). How will this scale the count of threads?

src/agentcore_rl_toolkit/client.py
```diff
@@ -172,8 +229,13 @@ def result(self, timeout: float = None) -> dict:
...
     """Client for invoking agents and collecting rollouts with full lifecycle management.
 
+    Provides both sync and async APIs. Sync methods (``invoke``, ``run_batch``)
+    block the caller. Async methods (``invoke_async``, ``run_batch_async``) are
+    suitable for use inside ``asyncio`` event loops.
+
     Note:
-        This client is NOT thread-safe. Use separate client instances for
+        This client is NOT thread-safe. Do not mix sync and async calls
+        concurrently on the same instance. Use separate client instances for
```
hellodanylo, line 244: Mostly libraries that I have seen typically use a separate Python class for async API.

src/agentcore_rl_toolkit/client.py
```diff
@@ -291,6 +355,50 @@ def _rate_limited_invoke(self, payload: dict, session_id: str, input_id: str) ->
...
+            self._async_lock = asyncio.Lock()
+        return self._async_lock
+
+    async def _async_rate_limited_invoke(self, payload: dict, session_id: str, input_id: str) -> RolloutFuture:
+        """Invoke with async TPS rate limiting.
+
+        The lock is held only during the timing check and released before the
+        HTTP call, so cold starts on one request don't block submission of others.
+        """
+        async with self._get_async_lock():
```
hellodanylo, line 382: For future, I recommend to refactor the rate limiting functionality into a reusable function. Here it's a little too tightly coupled with the rest of the ACR logic. For example, [backoff](https://github.com/litl/backoff) library has a nice, reusable interface.

src/agentcore_rl_toolkit/client.py
```diff
@@ -291,6 +355,50 @@ def _rate_limited_invoke(self, payload: dict, session_id: str, input_id: str) ->
...
+            response = self.agentcore_client.invoke_agent_runtime(
+                agentRuntimeArn=self.agent_runtime_arn,
+                runtimeSessionId=session_id,
+                payload=json.dumps(full_payload),
+            )
+            invoke_elapsed = time.time() - invoke_start
+            logger.info(f"Invoked session {session_id[:8]}... in {invoke_elapsed:.1f}s")
+            return self._parse_response(response)
+
+        data = await asyncio.to_thread(_invoke_and_parse)
```
hellodanylo, line 403: Same question as above about the count of threads created.

src/agentcore_rl_toolkit/client.py
```diff
@@ -465,3 +628,114 @@ def __iter__(self):
...
+                idx, start = submitting.pop(task)
+                try:
+                    future = task.result()
+                    future._poll_interval = self.initial_interval
+                    future._initial_interval = self.initial_interval
+                    future._max_interval = self.max_interval
+                    future._backoff_factor = self.backoff_factor
+                    active_futures[future.result_key] = (idx, future)
+                except Exception as e:
+                    yield BatchItem(success=False, error=str(e), index=idx, elapsed=time.time() - start)
```
hellodanylo: It might be better to format the exception like this:
```
def exc_to_full_string(exc: BaseException) -> str:
    return "".join(traceback.TracebackException.from_exception(exc).format())
```
Otherwise, it will miss stack trace and whatnot.

src/agentcore_rl_toolkit/client.py
```diff
@@ -465,3 +628,114 @@ def __iter__(self):
...
+
+            # Poll active futures that are ready and check for timeouts
+            completed_keys = []
+            for key, (idx, future) in active_futures.items():
+                if future.elapsed() > self.timeout:
+                    completed_keys.append(key)
+                    await asyncio.to_thread(future.cancel)
```
hellodanylo, line 723: Please comment on the count of threads created here.

src/agentcore_rl_toolkit/client.py
```diff
@@ -465,3 +628,114 @@ def __iter__(self):
...
+                        index=idx,
+                        elapsed=future.elapsed(),
+                    )
+                elif future.ready_to_poll() and await future.done_async():
+                    completed_keys.append(key)
+                    try:
+                        result = await future.result_async()
+                        yield BatchItem(success=True, result=result, index=idx, elapsed=future.elapsed())
+                    except Exception as e:
+                        yield BatchItem(success=False, error=str(e), index=idx, elapsed=future.elapsed())
```
hellodanylo: Same as above, with regards to formatting the exception as a string.

hellodanylo (commented):
Nice work with async futures! My only blocking concern is about the threads usage. I would like to see either a statement about how it will scale okay, or a fix to reduce the excessive thread creation.

src/agentcore_rl_toolkit/client.py
lyzustc, line 329: @luyuzhe111 We need to enable passing a base_url as input because different RL training examples could be routed to different vllm servers. Either make base_url a function input argument or allow reading it from payload.

src/agentcore_rl_toolkit/client.py
luyuzhe111: did some research. it looks like asyncio.to_thread uses loop.run_in_executor(None, ...) under the hood, which routes to Python's default executor — a ThreadPoolExecutor(max_workers=min(32, os.cpu_count() + 4)). So thread count is already bounded at ~32 regardless of how many concurrent futures exist.

src/agentcore_rl_toolkit/client.py
luyuzhe111: agreed. will do in a follow-up PR!

src/agentcore_rl_toolkit/client.py
luyuzhe111: @lyzustc fixed in the latest commit. thanks for the catch!!

src/agentcore_rl_toolkit/client.py
luyuzhe111: feel like separating them will result in some duplicated code. we can consider this route if async functionalities expand further?

PR #30, opened 2026-03-09
fix(client, migration-agent): add max_pool_connections to boto config, enable passing sampling parameters to evaluation script, polish migration agent example readme

lyzustc:
- Avoid throwing "Connection pool is full, discarding connection" warnings by setting max_pool_connections in boto config.
- Enable passing sampling parameters in evaluation script.
- Polish migration agent example's readme.

### DIFF
--- a/examples/strands_migration_agent/README.md
+++ b/examples/strands_migration_agent/README.md
@@ -78,6 +78,8 @@ mvn --version
 
 ## Installation
 
+Set the repo root and migration agent paths — these variables are used throughout this document:
+
 ```bash
 export TOOLKIT_ROOT=/path/to/your/agentcore-rl-toolkit/repo
 export MIGRATION_DIR=$TOOLKIT_ROOT/examples/strands_migration_agent
@@ -156,12 +158,9 @@ curl -X POST http://localhost:8080/invocations \
 
 ### Build & run locally
 
-Set the repo root path and migration agent path first, then build docker:
+Build the docker image:
 
 ```bash
-export TOOLKIT_ROOT=/path/to/your/agentcore-rl-toolkit/repo
-export MIGRATION_DIR=$TOOLKIT_ROOT/examples/strands_migration_agent
-
 docker buildx build \
   --build-context toolkit=$TOOLKIT_ROOT \
   -t migration:dev --load \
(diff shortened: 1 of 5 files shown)


examples/strands_migration_agent/README.md
```diff
@@ -141,42 +141,135 @@ curl -X POST http://localhost:8080/invocations \
...
+# Run with host network so the agent can access the locally hosted vLLM server
+docker run --network host --env-file $MIGRATION_DIR/.env migration:dev python -m dev_app
+
+# Submit request (same curl as above)
 ```
 
 ### Build & push to ECR
 
 ```bash
+cd /path/to/your/agentcore-rl-toolkit/repo
```
luyuzhe111: should we reuse TOOLKIT_DIR defined above? also applicable to some similar cd instructions later.

examples/strands_migration_agent/evaluate_async.py
```diff
@@ -240,13 +247,26 @@ async def main():
...
+    if args.sampling_params:
+        if isinstance(args.sampling_params, str):
+            sampling_params = json.loads(args.sampling_params)
+        else:
+            sampling_params = dict(args.sampling_params)
+
     # Create client
+    # In batch mode, concurrency is capped by max_concurrent.
+    # In individual mode, all payloads are submitted at once.
+    max_pool = args.max_concurrent if args.mode == "batch" else len(payloads)
```
luyuzhe111, line 261: hi youzhi, i actually also saw these warnings. after some investigation, i learned that those warnings actually don't suggest any issues. essentially boto3 will add new connections to the connection pool when the number of connections exceed the pool size. as the request (request submission) finishes, those extra connections boto3 will then be closed, leading to some warnings. 

since the cold start time for agentcore runtime will only improve, I feel like creating a lot of connections in the pool when each connection is only used less than a second can be wasteful. 

i still think exposing this option to set max_pool size is good, but i would suggest defaulting to 10 (boto3 default) and not infer it. especially when len(payloads) is really large (tens of thousands), we might run into problems. so in general i think we should delegate this responsibility to users when they know what they are doing.

examples/strands_migration_agent/README.md
lyzustc: Yeah, addressed in new commits.

luyuzhe111 (approved):

PR #31, opened 2026-03-10
feat(migration-agent): enable user to tune max_pool_connections

lyzustc:
max_pool_connections is not auto inferred in evaluation script, but user could set it.

### DIFF
--- a/examples/strands_migration_agent/README.md
+++ b/examples/strands_migration_agent/README.md
@@ -292,4 +292,12 @@ Note that all arguments can also be passed via CLI to override `config.toml` val
 
 Both `evaluate.py` (sync) and `evaluate_async.py` (async) can run multiple agent instances concurrently via `--max_concurrent`. The difference is that the sync script submits requests sequentially — a slow submission (e.g., ACR cold start) blocks the next one — while the async script dispatches submissions as concurrent tasks so cold starts don't block each other.
 
+#### Connection pool sizing
+
+Both scripts accept `--max_pool_connections` (default: 10) to control the urllib3 connection pool size for boto3 clients. If this value is smaller than `--max_concurrent`, you may see urllib3 warnings like `"Connection pool is full, discarding connection"`. This is **not an error** — requests still succeed, but excess connections are created and discarded instead of being reused from the pool, adding minor TCP/TLS handshake overhead. If you want to eliminate these warnings, you can set `--max_pool_connections` to match `--max_concurrent`:
+
+```bash
+python evaluate.py --exp_id my_eval --max_concurrent 50 --max_pool_connections 50
+```
+
 Results are saved as JSONL files under `results/` (e.g., `results/my_eval.jsonl`).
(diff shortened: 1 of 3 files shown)


luyuzhe111 (commented):
@lyzustc great work! I have a few high-level comments.

1. it would be great if you could split these different features into different PRs so it's easier for both human and coding agents to inspect the commit history. you can probably apply git diff on a feature branch (instead of main), make three commits, and cherry-pick them to different branches to submit the PRs. I think it's worth 4 PRs (max_pool_connection, auto-deploy, auto-cancellation, and async lock update). happy to review them. 

2. For on-the-fly deployment you added, previously I deliberately forced this deploy-then-run structure so that users test out their agent before running it. I guess having this support would be convenient in some cases but on the training framework side i think it's best we just expose agent_arn and s3 bucket, so that 1) we don't introduce a lot of extra args for deployment, and 2) people are forced to test out this agent deployment first so they don't hit a ton of rollout errors after the training engine sets up? what do you think?

3. for behavior changes please add some unit tests please (excluding the max pool connection). thanks!

examples/strands_migration_agent/README.md
```diff
@@ -292,4 +292,12 @@ Note that all arguments can also be passed via CLI to override `config.toml` val
 
 Both `evaluate.py` (sync) and `evaluate_async.py` (async) can run multiple agent instances concurrently via `--max_concurrent`. The difference is that the sync script submits requests sequentially — a slow submission (e.g., ACR cold start) blocks the next one — while the async script dispatches submissions as concurrent tasks so cold starts don't block each other.
 
+#### Connection pool sizing
```
luyuzhe111, line 295: good note

src/agentcore_rl_toolkit/client.py
```diff
@@ -318,6 +404,81 @@ def __init__(
...
+        if not image_uri:
+            missing.append("image_uri")
+        if not execution_role_arn:
+            missing.append("execution_role_arn")
+        if missing:
+            raise ValueError(
+                f"agent_runtime_arn is empty, so deployment config is required. " f"Missing: {', '.join(missing)}"
+            )
+
+        from bedrock_agentcore_starter_toolkit.services.runtime import BedrockAgentCoreClient
```
luyuzhe111: import at top since `bedrock_agentcore_starter_toolkit` is a required dependency anyway?

src/agentcore_rl_toolkit/client.py
```diff
@@ -373,9 +534,15 @@ def _rate_limited_invoke(self, payload: dict, session_id: str, input_id: str, **
         )
 
     def _get_async_lock(self) -> asyncio.Lock:
```
luyuzhe111, line 375: feel like this is worth a separate PR with additional unit tests.

src/agentcore_rl_toolkit/client.py
lyzustc: yeah will change it later.

luyuzhe111 (approved):

PR #38, opened 2026-03-16
feat(example): add Strands AppWorld agent example

luyuzhe111:
*Description of changes:*

Example Strands agent interacting with the AppWorld environment.


examples/strands_appworld_agent/rl_app.py
```diff
@@ -1,11 +1,10 @@
+import json
 import logging
 
-import requests
 from appworld.environment import AppWorld, AppWorldServers
-from mcp.client.streamable_http import streamable_http_client
 from reward import AppWorldReward
 from strands import Agent, tool
-from strands.tools.mcp import MCPClient
+from strands.agent.conversation_manager import NullConversationManager
```
luyuzhe111: probably not needed?

luyuzhe111 (commented):
@lyzustc great work!

examples/strands_appworld_agent/rl_app.py
lyzustc: @luyuzhe111 This is used in creating agent instance to prevent truncating conversation history (default behavior in strands agent using SlidingWindowConversationManager). I see migration and officebench examples both use it. For agent examples, if we do not want strands to do default context management we have to use it.

luyuzhe111: @lyzustc yea understood. originally I have this in migration agent because verl does not support per-turn training, so I had to make sure the conversation history is linear. that's why I had nullcontextmanager there. since the per-turn training is now supported, i don't think it's needed anymore? i was actually planning to remove it from the migration bench example. officebench example probably has that because claude saw the pattern. 

also wondering if the appworld agent is consistently hitting context limit? if not, then context management probably won't even be triggered?

lyzustc: @luyuzhe111 I removed NullConversationManager. I estimate around 20% of examples will hit context limit if using Qwen 3 30B Instruct model. But the impacts on outocme reward is slight.

lyzustc (approved):

PR #42, opened 2026-03-26
Sync migration agent

Linbo-Liu:
*Issue #, if available:*

*Description of changes:*
- Add max migration evaluation mode in `eval_util.py`, `evaluate.py`, and `evaluate_async.py`.
- Dependency changes in `pyproject.toml`
- Add reference to `README.md`.

### DIFF
--- a/examples/strands_migration_agent/evaluate_async.py
+++ b/examples/strands_migration_agent/evaluate_async.py
@@ -224,11 +224,16 @@ async def main():
         help="Whether a repository is evaluated under maximal migration",
     )
     parser.add_argument(
-        "--prompt_type",
-        type=str,
-        default="baseline",
-        choices=["baseline", "rag", "hybrid"],
-        help="Specify Java migration prompt type",
+        "--apply_static_update",
+        action="store_true",
+        default=False,
+        help="Whether to apply static update on JDK and dependency versions",
+    )
+    parser.add_argument(
+        "--use_dependency_search_tool",
+        action="store_true",
+        default=False,
+        help="Whether to allow dependency search tool for agent",
     )
 
     args = parser.parse_args()
@@ -255,7 +260,10 @@ async def main():
     logger.info(f"Found {len(s3_folder_uris)} repositories to evaluate")
 
     # Prepare payloads
-    payloads = [prepare_payload(uri, args.require_maximal_migration, args.prompt_type) for uri in s3_folder_uris]
+    payloads = [
+        prepare_payload(uri, args.require_maximal_migration, args.apply_static_update, args.use_dependency_search_tool)
+        for uri in s3_folder_uris
+    ]
 
     # Setup results directory and file
     results_dir = Path(__file__).parent / "results"
--- a/examples/strands_migration_agent/evaluate_async.py
+++ b/examples/strands_migration_agent/evaluate_async.py
@@ -7,12 +7,7 @@ import logging
 import time
 from pathlib import Path
 
-from eval_utils import (
-    append_result_to_file,
-    get_s3_folder_uris,
-    load_config,
-    prepare_payload,
-)
+from eval_utils import append_result_to_file, get_s3_folder_uris, load_config, prepare_payload
 
 from agentcore_rl_toolkit import RolloutClient
 
@@ -229,11 +224,11 @@ async def main():
         help="Whether a repository is evaluated under maximal migration",
     )
     parser.add_argument(
-        "--agent_type",
+        "--prompt_type",
         type=str,
         default="baseline",
         choices=["baseline", "rag", "hybrid"],
-        help="Specify Java migration agent type",
+        help="Specify Java migration prompt type",
     )
 
     args = parser.parse_args()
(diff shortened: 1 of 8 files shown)


lyzustc (approved):

examples/strands_migration_agent/rl_app.py
```diff
@@ -63,6 +99,9 @@ def invoke_agent(payload: dict):
 
     start_time = time.time()
     setup_repo_environment(repo_path)
+    if agent_type == "hybrid":
```
luyuzhe111: nit: might be cleaner if factoring these two calls into `setup_repo_environment()`?

luyuzhe111 (commented):
@Linbo-Liu you might need to install the pre-commit hooks properly. a lot of the cosmetic wraps seem to stem from using the default 88-char ruff line limit instead of the 120-char limit specified in this repo.

examples/strands_migration_agent/rl_app.py
```diff
@@ -43,17 +44,48 @@
...
-    model = OpenAIModel(client_args={"api_key": "EMPTY", "base_url": base_url}, model_id=model_id, params=params)
+    request = InvocationRequest(**payload)
+    prompt = system_prompt
+    if request.require_maximal_migration:
+        prompt += (
+            "\nYou should update all dependencies in the `pom.xml` file to their latest versions that support Java 17."
+        )
+    if agent_type == "rag":
+        prompt += (
+            "You have access to a dependency version lookup tool. When updating dependencies "
```
luyuzhe111: do we want a "\n" here as well? or do you intend to directly follow the max migration prompt (in which case a empty space will be needed?)

examples/strands_migration_agent/rl_app.py
```diff
@@ -43,17 +44,48 @@
...
+            "in pom.xml:\n"
+            "1. Use the search_dependency_version tool to look up the recommended Java 17 "
+            "compatible version for each dependency\n"
+            "2. If a dependency is not found in the database, use your knowledge to select "
+            "an appropriate version\n"
+            "3. Update all dependencies to their Java 17 compatible versions"
+        )
+        tools.append(search_dependency_version)
+    elif agent_type == "hybrid":
```
luyuzhe111: wondering what's the interaction between agent type and `require_maximal_migration`? The hybrid type calls update_jdk_related() + update_dependency_version() (which presumably does maximal migration preprocessing), but             require_maximal_migration is a separate boolean. can we make it more explicit?

luyuzhe111: based on my understanding, the hybrid agent type isn't really a different agent — it's a preprocessing toggle (apply static dependency updates before the agent runs). Making this an explicit boolean like apply_static_update would be clearer and composable with other flags like require_maximal_migration. Same argument could apply to rag — these feel like independent feature flags, not mutually    exclusive agent types. wondering what's your thought?

examples/strands_migration_agent/rl_app.py
Linbo-Liu: +1. Moved it into `setup_repo_environment`.

examples/strands_migration_agent/rl_app.py
Linbo-Liu: Good catch. yes. added `\n`.

examples/strands_migration_agent/rl_app.py
Linbo-Liu: You are right. If one uses hybrid agent, the implicit intention is to do max migration. However, I didn't set this explicit because technically speaking, if doing minimal migration, hybrid agent can downgrade the dependency versions even if they were already upgraded by static script...

Linbo-Liu: >hybrid agent type isn't really a different agent

It uses a different prompt. RAG agent is the same: different prompt. Maybe I should change `agent_type` to `prompt_type`?

luyuzhe111: @Linbo-Liu thanks for the explanation! it makes a lot of sense. 

in this case, I would suggest flattening it into two independent boolean flags: `apply_static_update`, and `use_dependency_search_tool`. It will justify the following changes better than a single `prompt type`. 

- for `setup_repo_environment`, it will take an additional arg, `apply_static_update`. it's semantically much clearer. with `prompt type` as input, people get confused how prompt type is related to the repo setup procedure. 
- it will also be more intuitive imo to update both the tool and prompt based on the `use_dependency_search_tool` flag rather than prompt type.

examples/strands_migration_agent/evaluate_async.py
```diff
@@ -7,7 +7,12 @@
 import time
 from pathlib import Path
 
-from eval_utils import append_result_to_file, get_s3_folder_uris, load_config, prepare_payload
+from eval_utils import (
```
luyuzhe111: probably remaining noise from previous default ruff?

examples/strands_migration_agent/rl_app.py
```diff
@@ -43,17 +44,46 @@
 def invoke_agent(payload: dict):
     base_url = payload["_rollout"]["base_url"]
     model_id = payload["_rollout"]["model_id"]
+    prompt_type = payload.get("prompt_type", "baseline")
     params = payload["_rollout"].get("sampling_params", {})
+    tools = [shell, editor]
+
+    request = InvocationRequest(**payload)
+    prompt = system_prompt
```
luyuzhe111, line 51: nit: system_prompt is not used later? in this case, we can just do system_prompt += "xxx"?

Linbo-Liu: Tried it but will have local variable referenced before assignment error.

luyuzhe111 (approved):

luyuzhe111:
thanks for the great work and the iterations!

PR #51, opened 2026-04-06
fix(client): use a unified rate limiter for all ACR API calls

lyzustc:
Use a unified rate limiter for all ACR API calls, this avoids throttle errors in ACR session cancellation.

### DIFF
--- a/src/agentcore_rl_toolkit/client.py
+++ b/src/agentcore_rl_toolkit/client.py
@@ -21,6 +21,60 @@ def _format_exception(exc: BaseException) -> str:
     return "".join(traceback.format_exception(exc))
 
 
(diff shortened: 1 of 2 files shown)


luyuzhe111:
@lyzustc this is a great improvement! does stop session API and invoke agent API share the same rate limit tho?

lyzustc:
> @lyzustc this is a great improvement! does stop session API and invoke agent API share the same rate limit tho?

I actually don't know if ACR internally uses a shared rate limit. This implementation safely prevents many sessions timeout simultaneously -> too many stop session requests cause throttle.

lliquid (approved):

PR #58, opened 2026-05-04
feat(migration): Enable using AWS CodeArtifact as maven mirror in Mig…

lyzustc:
Migration agent sessions at Bedrock AgentCore Runtime could fail because making too many requests to public maven source for downloading Java dependencies. Add an resolving solution of using AWS CodeArtifact as maven mirror source to cache the downloaded Jave dependencies.


Linbo-Liu (approved):
LGTM.

PR #59, opened 2026-05-04
docs(roadmap): core-api rename, SlimeRunner plans committed; slime data-contract done

lliquid:
## Summary

Tracks design plans in \`docs/roadmap/\`, split into \`committed/\` (approved, not yet implemented) and \`done/\` (shipped).

**What's tracked:**
- \`docs/roadmap/committed/core-api-rename.md\` — approved plan for the 0.2.0 rename (\`app.py\` / \`client.py\` / \`reward_function.py\` → \`rollout_executor.py\` / \`rollout_launcher.py\` / \`reward.py\`; \`RolloutClient\` → \`RolloutLauncher\`; \`payload["_rollout"]\` → \`payload["rollout_config"]\`; no shims, single PR when implemented).
- \`docs/roadmap/committed/slime-runner-entrypoint.md\` — approved plan to wrap \`train.sh\` + \`config.yaml\` behind a single \`SlimeRunner\` Python class as the primary entry point; \`train.sh\` stays as the low-level escape hatch.
- \`docs/roadmap/done/slime-data-contract.md\` — shipped in #61. Makes the JSONL row's \`metadata\` dict the agent payload verbatim, so non-GSM8K agents (migration / AppWorld / OfficeBench) train without slime-integration changes. Appendix A documents how slime's \`group_index\` is assigned, for future readers.
- \`.gitignore\` updates so drafts (\`docs/roadmap/draft/\`), PR-body scratchpads (\`docs/pr/\`), and personal tooling (\`docs/superpowers/\`) stay untracked.

**Not tracked:** roadmap drafts stay local until they're promoted to \`committed/\`; \`done/\` plans are committed after their implementation PR merges.

## Plan scope at a glance

- **core-api-rename:** one PR, one version bump (0.2.0). No deprecation scaffolding — at this adoption stage the carrying cost outweighs the migration burden. Three files rename, four classes rename, one payload key flips. Backend impact: slime (in-tree, bundled), rllm (out-of-tree, tracking issue only), verl (not integrated yet).
- **slime-runner-entrypoint:** additive convenience layer. \`SlimeRunner(...).train()\` replaces the train.sh + config.yaml + env-var combo. \`train.sh\` stays untouched.
- **slime-data-contract (done):** implemented in #61. End-to-end smoke test verified reward climbs 0.27 → 0.63 over 10 rollouts on Qwen2.5-3B + GSM8K.

## Test plan

- [ ] This is a roadmap/documentation PR — no code changes, nothing to functionally test.
- [ ] Reviewers confirm each plan reads clearly and the scope is correct.


docs/roadmap/committed/core-api-rename.md
```diff
@@ -0,0 +1,232 @@
...
+`AgentCoreRLApp`, `RolloutFuture`, and `RewardFunction` stay
+unchanged.
+
+### Changes at a glance
+
+**Module files:**
+
+| Today | Renamed |
+|---|---|
+| `app.py` | `rollout_server.py` |
```
lyzustc: I think it is better to rename `app.py` to `rollout_executor.py`? Conceptually `AgentCoreRLApp` class does not start a server that handles multiple rollout requests, but packs agent codes into an app that will be executed in every ACR session.

lyzustc (commented):
I feel renaming `app.py` to `rollout_server.py` is not appropriate, other things look good to me.

docs/roadmap/committed/core-api-rename.md
lliquid: good idea.

lliquid:
> I feel renaming `app.py` to `rollout_server.py` is not appropriate, other things look good to me.

revised the PR , pls check

docs/roadmap/committed/core-api-rename.md
luyuzhe111: wait agentcore rl app does start a server?

lyzustc: @luyuzhe111 I think if naming to rollout_server, people may easily misunderstand it works like an agent server --- handle input requests and return full agent rollout traces, but actually it does not, rollout traces are collected from model gateway.

luyuzhe111: makes sense. so the `app.py` naming comes from [here](https://github.com/aws/bedrock-agentcore-sdk-python/blob/main/src/bedrock_agentcore/runtime/app.py), which holds the server app agentcore rl app inherits.

luyuzhe111: @lliquid so for me personally i would just keep it as is due to 1) similar bedrock sdk convention, and 2) `app.py` is not really public-facing as users import the `AgentCoreRLApp` directly, not via an `app` module.

PR #72, opened 2026-06-03
feat(slime): add sglang patch for v0.5.12.post1.

Linbo-Liu:
*Issue #, if available:*
The latest `slimerl/slime:latest` image uses sglang `v0.5.12.post1`. The sglang patch for v0.5.9 `sglang_token_ids.py` can't be successfully applied due to changes to sglang. 

*Description of changes:*
Add a patch compatible with sglang v0.5.12.post1 in `sglang_token_ids_v_0_5_12.py`.

Tested via `verify_sglang_token_ids.py` with sglang v0.5.12.post1 in `slimerl/slime:latest` docker container. Test output:
```
[1/4] non_streaming_basic            PASS
[2/4] streaming_basic                PASS
[3/4] cross_mode_consistency         PASS
[4/4] tool_call_consistency          PASS

OK: 4/4 checks passed
```

### DIFF
--- a/src/agentcore_rl_toolkit/backends/slime/patches/sglang_token_ids_v_0_5_12.py
+++ b/src/agentcore_rl_toolkit/backends/slime/patches/sglang_token_ids_v_0_5_12.py
@@ -1,0 +1,837 @@
+"""Apply SGLang v0.5.12.post1 patch to expose prompt/completion token IDs in chat completions.
+
+Same goal as sglang_token_ids.py (v0.5.9) but for the refactored v0.5.12 streaming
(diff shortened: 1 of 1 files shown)


luyuzhe111:
@Linbo-Liu thanks for the PR! I looked into slime's mechanism to preserve prompt & response IDs. https://github.com/THUDM/slime/blob/main/slime/agent/adapters/openai.py. Since we will install slime anyway, could you investigate if there's something we can reuse, potentially in place of rllm-model-gateway?

lyzustc:
Close as we are planning to use sglang /generate API for rollout endpoint in gateway, no need to do sglang patch for token id capture.

PR #78, opened 2026-07-01
feat(slime): multi-turn trace merging, validation at an input dataset and gateway config fix in custom rollout function

lyzustc:
This PR adds drift-free multi-turn trajectory merging to the slime custom rollout
function and fixes several issues that surfaced when running on full training /
validation datasets (eval episodes timing out and crashing the engine, connection-pool
churn, and unreadably noisy logs). It also brings `SlimeRunner` and the example scripts
in line with these changes.

## Custom rollout function (`integration/`)

- **feat — auto-merge multi-turn conversations (`traces.py`).** Added
  `merge_traces_to_samples`: an episode's turns are folded into one training
  Sample per contiguous prefix-extending segment (bridge tokens masked out,
  completions trained), instead of one Sample per turn. Pairs with the gateway's
  cumulative token mode; falls back to one-Sample-per-turn when a turn breaks the
  prefix.
- **feat — cumulative token mode plumbing (`rollout.py`, `gateway.py`).** New
  `cumulative_token_mode` / `renderer_family` config knobs, forwarded to the
  gateway; `merge_traces_to_samples` is wired in as the rollout's sample builder.
- **feat — validation (`rollout.py`).**  Supported doing validation on an input validation
  set periodically with an interval in training. Bounded in-flight agent sessions in validation 
  by `max_concurrent` in yaml config. 
- **fix — connection-pool churn (`rollout.py`).** Added `max_pool_connections`
  so user can set `>= max_concurrent`to stop boto3 client logging "Connection pool
  is full, discarding connection" under concurrent ACR sessions.
- **fix — log noise (`rollout.py`, `gateway.py`).** Replaced per-session logging
  with a single per-batch summary (`episodes / succeeded / failed / sequences`);
  failed episodes log once at INFO with the full ACR `session_id` for CloudWatch
  lookup. Added `gateway_log_level` (default `warning`) to silence the gateway's
  uvicorn/httpx access logs.
- **fix — gateway always uses SGLang `/generate` (`gateway.py`).** The slime
  backend now enables `use_sglang` unconditionally and always passes the served
  `--model`; removed the dead `use_sglang` config field. (The `/generate`
  mechanics live in the rllm-model-gateway PR.)

## Training scripts & runner

- **`runner.py`** — exposed the new knobs on `SlimeRunner`
  (`sglang_tool_call_parser`, `sglang_reasoning_parser`, `cumulative_token_mode`,
  `renderer_family`, `max_pool_connections`, `gateway_log_level`,
  `sglang_context_length`) and added `cuda_home` to pin CUDA_HOME/LD_LIBRARY_PATH
  (incl. `--train-env-vars` for the Megatron actors), fixing TransformerEngine's
  "Multiple libcudart" abort. The tool-call parser is now a field instead of a
  hardcoded `qwen25`.
- **`examples/math_agent/train.sh`** — ported the CUDA-toolchain pinning,
  `torch_memory_saver` fixup, eval flags, log suppression, and checkpointing from
  the working run; all paths/credentials stay env-var placeholders.
- **`config.yaml.example`** — documented the new `max_pool_connections`,
  `cumulative_token_mode`, and `renderer_family` settings.

This PR must work together with https://github.com/rllm-org/rllm/pull/715.
A follow-up PR with slime environment setup guide in docs changes will come soon.


src/agentcore_rl_toolkit/backends/slime/examples/math_agent/config.yaml.example
```diff
@@ -17,5 +17,19 @@ gateway_port: 9090                  # rllm-model-gateway port
...
+# sequence and forwards it to sglang servers if the message history is cumulative.
+cumulative_token_mode: false
+# Model gateway's cumulative token mode uses renderers package to construct
+# cross-message-turn bridge tokens, that needs setting model family name
+# (e.g. "qwen3", "glm-5"). If it is "auto" and actor_rollout_ref.model.path
+# is a huggingface model id, renderers will auto infers model family name.
+# But you must set model family name explicitly if actor_rollout_ref.model.path
+# is a local model path. Check supported model families in MODEL_RENDERER_MAP of
+# https://github.com/PrimeIntellect-ai/renderers/blob/main/renderers/base.py
+renderer_family: "auto"
```
lliquid: the semantics of renderer family does not look clear - if i am a user of the trainer class. if this could be inferred from model name then probably do not need to expose this hyperparam ?

lyzustc: Thanks for feedback. Renderer family cannot be inferred from local model path, in this case user must explicitly set it. I changed the comments here to clarify.

src/agentcore_rl_toolkit/backends/slime/examples/math_agent/config.yaml.example
```diff
@@ -17,5 +17,25 @@ gateway_port: 9090                  # rllm-model-gateway port
 acr_timeout: 900                    # per-session ACR invocation timeout (s)
 model_id: "default"                 # OpenAI model id served to the agent
 acr_tps_limit: 25                   # ACR service TPS quota
-max_concurrent: 100                 # max concurrent ACR sessions (eval batching)
+max_concurrent: 100                 # max concurrent ACR sessions
+max_pool_connections: 256           # boto3 connection pool size
```
lliquid: it feels like max_concurrent and max_pool_connections are controlling similar things - do we need to specify these two arguments or the effective one is actually the one with lower value ?

lyzustc: No, they are different. max_pool_connections are pased to RolloutClient initialize here https://github.com/awslabs/agentcore-rl-toolkit/blob/main/src/agentcore_rl_toolkit/client.py#L387, controlling boto3 connection bool size, not the max number of concurrently running ACR agent sessions.

src/agentcore_rl_toolkit/backends/slime/runner.py
```diff
@@ -152,17 +172,41 @@ def _source_model_script(self) -> list[str]:
         items = out.split(b"\0")
         return [x.decode() for x in items if x]
 
+    def _cuda_ld_library_path(self) -> str:
```
luyuzhe111: do we really need this at the runner level? how did slime handle this?

lyzustc: Good question, slime handles this the same way — its example launch scripts set `LD_LIBRARY_PATH` inline in the ray job submit --runtime-env-json env-vars block (e.g. [scripts/run-deepseek-r1.sh](https://github.com/THUDM/slime/blob/main/scripts/run-deepseek-r1.sh#L165) hardcodes `LD_LIBRARY_PATH`). SlimeRunner just populates that same worker runtime_env, so it's consistent with slime's approach. But we compute the value instead of hardcoding it, because slime's scripts assume using the fixed docker with cu12.9 while we allow the user to run in arbitrary venvs where different cuda versions (12.9 or 13.0) could be installed. It's harmless and opt-in: the helper only runs when cuda_home is set. Left unset (the default), SlimeRunner inherits the ambient env with no pinning — identical to slime's default behavior.

luyuzhe111: sg

PR #88, opened 2026-07-20
feat: add sandbox SDK (sync client + sandboxd + quickstart example)

luyuzhe111:
## Motivation

Coding-agent evaluation and RL training run against large collections of prebuilt Docker images (e.g. SWE-bench-style environments): a filesystem, dependencies, and a test runner waiting for shell commands — not HTTP agent servers. AgentCore Runtime already has the right primitives for this workload (per-session microVM isolation, session routing, auto-scaling, and a native `InvokeAgentRuntimeCommand` API that executes shell commands inside a live session), but such images cannot be deployed today: ACR requires containers to expose `/ping` and `/invocations` on port 8080, and session health/busy state must be managed correctly or sessions get idle-reaped mid-task.

This PR closes that gap so any image with a shell can serve as a scalable, isolated command-execution sandbox on ACR.

## Approach

Two small pieces, deliberately kept apart:

1. **`sandboxd/` — a minimal Go health shim** (stdlib-only, single static binary) added to the image as its entrypoint. It satisfies the ACR container contract and holds one busy/healthy flag: `/invocations {"action":"start"}` flips ping to `HealthyBusy` (session survives idle timeout across quiet gaps between commands), `{"action":"stop"}` flips it back. It contains **no execution logic**; unknown JSON fields are ignored for forward compatibility.

2. **`agentcore_rl_toolkit.sandbox` — a sync Python client** (boto3 only, zero new deps). `client.start()` / `client.attach(session_id)` / `sb.exec(...)` / `sb.terminate()`. Command execution goes through ACR's native `InvokeAgentRuntimeCommand`, never through the shim; the client consumes the event stream into `ExecResult(exit_code, stdout, stderr, timed_out)`.

Key semantics:

- **Failure is data**: nonzero exit codes and timeouts are ordinary results (the reward signal in eval/RL), never exceptions — exceptions are reserved for infrastructure faults (`ClientError`, `EventStreamError`, `SandboxProtocolError` for misdeployed images).
- **Stateless commands**: each `exec()` runs in a fresh process; `cwd=`/`env=` are composed into the command per call. The command API doesn't invoke a shell itself, so the client wraps commands in `<shell> -c '...'` (default `/bin/sh` for arbitrary-image portability, overridable).
- **Fault-tolerant teardown**: `terminate()` flips ping to Healthy *before* calling `StopRuntimeSession`, so the idle reaper collects the session even if the stop call fails; idempotent, safe in `__exit__`.

`examples/sandbox_quickstart/` provides the end-to-end walkthrough: build the binary (`sandboxd/build.sh`, cross-compiles arm64 on x86 hosts with a golang-container fallback), wrap + push image, create the runtime, run commands. The sandbox image does not contain this toolkit — the SDK is purely client-side.

## Scope

First PR of a phased feature. Deliberately excluded (planned next): async client (aiobotocore), TTL-based leak protection, image wrapping/ECR automation in the SDK, detached exec, interactive shells, file transfer.

## Testing

- `tests/sandbox/`: unit tests with fully mocked boto3 (fake event streams) — no AWS access needed; runs in existing `unit-tests.yml`.
- `sandboxd/`: Go unit tests incl. race detector; new path-gated CI workflow (`.github/workflows/sandboxd.yml`: gofmt, vet, build, `go test -race`).
- Shell-wrapping and exec semantics verified against the live service; quickstart validated end-to-end on a deployed runtime (`exit_code=0`, `hello from aarch64`).

### DIFF
--- a/src/agentcore_rl_toolkit/sandbox/client.py
+++ b/src/agentcore_rl_toolkit/sandbox/client.py
@@ -1,0 +1,352 @@
+"""Sync client for running shell commands in AgentCore Runtime sandbox sessions.
+
+A "sandbox" is an AgentCore Runtime session whose container runs the
+``agentcore-sandboxd`` health shim (see ``sandboxd/`` at the repo root). The
+shim only manages the Healthy/HealthyBusy ping state; command execution goes
+through AgentCore Runtime's native ``InvokeAgentRuntimeCommand`` API, which runs
+shell commands inside the same session/container.
+
+Usage:
+    from agentcore_rl_toolkit.sandbox import SandboxClient
+
+    client = SandboxClient(runtime_arn="arn:aws:bedrock-agentcore:...")
+    with client.start() as sb:
+        result = sb.exec("cd /app && pytest -q", timeout=900)
+        print(result.exit_code, result.stdout)
+"""
+
+import json
+import logging
+import re
+import shlex
+import uuid
+
+import boto3
+from botocore.config import Config
+
+from .types import ExecResult, SandboxProtocolError
+
+logger = logging.getLogger(__name__)
+
+# runtimeSessionId constraints from the bedrock-agentcore service model.
+_SESSION_ID_MIN_LEN = 33
+_SESSION_ID_MAX_LEN = 256
+
+_ENV_KEY_RE = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
+
+
+def _wrap_in_shell(command: str, shell: str = "/bin/sh") -> str:
+    """Wrap a shell command string for InvokeAgentRuntimeCommand.
+
+    The command API does NOT interpret the command through a shell — it
+    word-splits argv-style, passing ``;``, ``|``, ``$VAR`` etc. through as
+    literal arguments (verified against the live service; every example in the
+    official docs likewise wraps commands, in ``/bin/bash -c "..."``:
+    https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/runtime-execute-command.html).
+    Shell semantics therefore require an explicit ``<shell> -c`` wrapper.
+
+    We default to ``/bin/sh`` rather than the docs' ``/bin/bash`` because the
+    sandbox use case targets arbitrary images: POSIX sh exists in any image
+    with a shell at all (including busybox/alpine), while bash is often absent
+    from minimal images.
+
+    Wrapper quoting, per the service tokenizer's semantics (live-verified):
+
+    - Single-quoted args pass through fully verbatim (backslashes untouched),
+      but POSIX quote concatenation (``'it'"'"'s'``) is NOT supported — so the
+      single-quote wrapper only fits commands without single quotes.
+    - Double-quoted args support ``\\"`` and ``\\\\`` escapes, and the tokenizer
+      does NOT expand ``$`` or backticks (they reach the inner shell, which
+      expands them — the semantics we want). Escaping ``\\`` and ``"`` therefore
+      round-trips any command exactly.
+
+    Both forms keep the wire format human-readable — the service logs the input
+    command to the runtime's CloudWatch log group for auditing.
+    """
+    if "'" not in command:
+        return f"{shell} -c '{command}'"
+    escaped = command.replace("\\", "\\\\").replace('"', '\\"')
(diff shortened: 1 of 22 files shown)


src/agentcore_rl_toolkit/sandbox/client.py
```diff
@@ -0,0 +1,355 @@
...
+        ``{"action": "stop"}`` (so that even if the subsequent stop call fails,
+        the idle reaper collects the session within the idle timeout), then call
+        ``StopRuntimeSession``. Failures are logged as warnings, never raised —
+        this must be safe to call from ``__exit__``.
+        """
+        if self._terminated:
+            return
+        self._terminated = True
+
+        try:
```
Teq2412, line 339: I'm thinking is it worthy to add a backoff retry here, given that if both try fail, it will take up to 8 hours for the ACR to forcely kill the MicroVM and give an unexpected cost ( small tho )

src/agentcore_rl_toolkit/sandbox/client.py
```diff
@@ -0,0 +1,355 @@
...
+        Returns:
+            An ``ExecResult`` with exit code, accumulated stdout/stderr, and the
+            timeout flag.
+        """
+        composed = _compose_command(command, cwd=cwd, env=env)
+        body = {"command": _wrap_in_shell(composed, shell=shell or self._client.shell)}
+        if timeout is not None:
+            body["timeout"] = timeout
+
+        response = self._client._client.invoke_agent_runtime_command(
```
Teq2412, line 287: Should we raise error when exec is called out of client.start() ? [when self._terminated == True]

tests/sandbox/test_client.py
```diff
@@ -0,0 +1,319 @@
...
+import io
+import json
+from unittest.mock import MagicMock, patch
+
+import pytest
+from botocore.exceptions import ClientError, EventStreamError
+
+from agentcore_rl_toolkit.sandbox import Sandbox, SandboxClient, SandboxProtocolError
+
+FAKE_ARN = "arn:aws:bedrock-agentcore:us-west-2:123456789012:runtime/sandbox-test"
```
lliquid, line 12: https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/runtime-get-started-cli.html#configure-agent 
@luyuzhe111  can we use the local agent cli set up for integration test?

lyzustc (approved):

lyzustc:
Nice work! I suggest to add introduction and usage guidance pages to docs site when sandbox SDK feature is complete.

tests/sandbox/test_client.py
luyuzhe111: @lliquid ah good point. in this PR, I used `examples/sandbox_quickstart/run_sandbox.py` to do local test, but this is manual and hard to reproduce. I can set up the integration test workflow in a follow-up PR so it can be triggered automatically. and yes we should be able to leverage the agentcore cli for this purpose.

src/agentcore_rl_toolkit/sandbox/client.py
luyuzhe111: @Teq2412 Good thinking about the zombie-cost case! Two mitigations are currently in place: (1) the boto3 client is built with Config(retries={"max_attempts": 5, "mode": "adaptive"}), so both calls in terminate() get exponential-backoff retries at the botocore layer; (2) the two steps are ordered so only one needs to succeed — the ping is flipped to Healthy first, so even if StopRuntimeSession fails after all retries, the idle reaper collects the session in ~15 min, not 8 h. The 8-hour worst case requires both calls to fail post-retry, which in practice means the client process died or lost connectivity — a case no in-process retry survives. That failure class is the motivation for the planned TTL watchdog in sandboxd (next phase): the start invocation carries a ttl_seconds and the server flips itself back to Healthy on expiry. Deferred out of this PR since at prototype scale a rare leaked session dying at maxLifetime is acceptable; the sandboxd protocol already ignores unknown fields so adding it later is non-breaking!

src/agentcore_rl_toolkit/sandbox/client.py
luyuzhe111: Great catch! fixed in the latest commit. exec() now raises RuntimeError on a terminated handle, with unit tests for both the explicit terminate() and context-manager-exit paths.

PR #91, opened 2026-07-26
feat(verl): experimental verl backend on the rollout gateway, validated via CI and e2e GSM8K run

luyuzhe111:
## What

A new experimental verl integration (`backends/experimental/verl/`) built on the in-repo rollout gateway, intended to succeed `backends/verl` (which stays untouched until this graduates). Validated end to end: GRPO full fine-tuning of Qwen3-4B against a live AgentCore runtime reaches **~0.93 GSM8K validation reward in one epoch** (reproduced
twice). Training curves/screenshots attached below.

Screenshot 2026-07-25 at 7 49 09 PM

## Why: a more modular integration

The existing `backends/verl` predates the rollout gateway and pays for it structurally: it depends on the external `rllm-model-gateway`, subclasses `RayPPOTrainer` with a custom`fit()` (upstream is removing `RayPPOTrainer`; the unified v1 trainer is now default), and replaces verl's entrypoint — cutting it off from verl-native features.

This backend inverts that: verl stays unchanged, and we plug in at the seams verl designed for external agent frameworks (its agent-loop module names AWS Bedrock AgentCore as an intended integration thanks to our [PR](https://github.com/verl-project/verl/pull/4216) last year). verl is pinned to `78bba31d` (a 0.9.0.dev commit — the one uni-agent, verl's official agentic layer, currently verifies via its submodule).

| | legacy `backends/verl` | this backend |
|---|---|---|
| Entrypoint | custom `main.py` + `AgentCoreTrainer(RayPPOTrainer)` | default `python -m verl.trainer.main_ppo` |
| Trainer | legacy RayPPOTrainer (deprecated upstream) | v1 trainer (`trainer.use_v1=true`) |
| Token capture | external rllm-model-gateway subprocess, per-session URLs | in-repo gateway, in-process; session id rides the api-key slot |
| Tokens | scraped from chat responses | token-in/token-out via verl's `LLMServerClient` |
| Multi-turn merging | trainer-side segment merge | inside the gateway (`TrajectoryManager`) |
| verl features | custom fit loop (GRPO/REMAX only) | native: replay buffer, checkpointing, validation, reward loop, rollout correction (TIS), async modes |

Each concern is one small module: `sampling_backend.py` (gateway `SamplingBackend` over verl's token-in/token-out client), `gateway_host.py` (one gateway per worker process), `agent_loop.py` (the rollout loop; failures never raise — inert rows keep the batch alive), `dataset.py` (`PayloadDataset`).

Two contracts keep agents decoupled from the trainer:

- **Dataset**: rows carry the agent's exact invoke payload in a single `payload` column; the chat-format `prompt` column verl's dataloader needs is synthesized at load time. Agents never learn dataset/trainer conventions, trainers never guess agent inputs.
- **Reward** (`reward_mode`): `built_in` — the agent returns `{"rewards": ...}` with its session result, which becomes `rm_scores` directly. Trainer-side scoring (`separate`) is rejected at startup for now: verl's reward managers index `data_source` and `reward_model.ground_truth` unguarded, which the payload-first dataset contract does not provide.

Agent-side change is one line (math example, validated; other examples migrate as they are validated): `api_key = context.session_id or "EMPTY"` — the ACR session id doubles as the trajectory-capture key, and `"EMPTY"` keeps local runs and the legacy per-session-URL path working.

## Tests & CI

- **Integration tests against real verl** (`tests/backends/experimental/verl/`, 45 tests): the loop is constructed through verl's own `AgentLoopBase`/hydra instantiation with a live threaded gateway and the real OpenAI SDK on the wire; only the LLM server client (fronts Ray + a GPU engine) and the tokenizer are faked. 
- **New CI workflow** (`experimental-verl-integration.yml`): runs those tests on GitHub runners against the pinned verl via a `verl-experimental-ci` extra (CPU torch, no inference engines — the tests don't need one), locked sync, path-filtered triggers. The plain unit-test job skips them (no verl in the dev extra).
- **E2E on live infrastructure**: full GRPO runs on 8 GPUs against a deployed AgentCore math agent — thousands of rollouts, zero failures; token capture verified byte-accurate (`rollout_actor_probs_pearson_corr ≈ 0.997`). Run screenshot attached.

## Install / run

```bash
uv sync --extra verl-experimental          # verl pinned via [tool.uv.sources]
cd src/agentcore_rl_toolkit/backends/experimental/verl/examples/math_agent
export AGENT_RUNTIME_ARN=... ACR_S3_BUCKET=... 
python preprocess_gsm8k.py --output-dir gsm8k
./fsdp_fft_sync_grpo.sh
```

See `backends/experimental/verl/README.md` for the design, config reference, and troubleshooting notes.

## Next Steps

- **Trainer-side rewards (`reward_mode="separate"`)**: needs a concrete reward function to validate a synthesized `data_source` / `reward_model.ground_truth` column shape against — picking one blind would just be a guess, so the mode stays closed until there is a real consumer.
- **Failure visibility in metrics**: failed rollouts are logged per rollout and tagged `acr_failed` in `extra_fields`, but nothing aggregates that into a training-step metric — a fully-failing run shows up only as `response_length/mean == 1.0` and flat rewards on a dashboard. verl's only generic `extra_fields` -> metrics path runs in `_validate()`, so a training-side failure-rate metric needs its own aggregation hook.


src/agentcore_rl_toolkit/backends/experimental/verl/examples/math_agent/agentcore_agent.yaml
Linbo-Liu, line 1: Do you want to create xxx.yaml.example files so that we won't commit private info into the yaml in later commits?

src/agentcore_rl_toolkit/backends/experimental/verl/agent_loop.py
```diff
@@ -0,0 +1,426 @@
...
+
+# NOT decorated with verl's @register: registration comes solely from the
+# agent_loop_config_path YAML entry, because that entry also carries the
+# constructor kwargs this loop requires (agent_runtime_arn, s3_bucket, ...) —
+# @register stores only a bare {"_target_": ...} with no kwargs, which is why
+# it suffices for verl's built-in loops but not here. Worse, the decorator
+# fires when hydra first imports this module (at the first instantiation) and
+# would overwrite the YAML's kwarg-carrying registry entry with the bare one:
+# the first rollout works, every subsequent one crashes on missing kwargs.
+class AgentCoreAgentLoop(AgentLoopBase):
```
Linbo-Liu, line 94: Noticed there is already AgentCoreLoopManager in backends/verl/loop_manager.py. To confirm, the reason for a new AgentCoreAgentLoop is that it's compatible with verl's new `AgentLoopBase`?

src/agentcore_rl_toolkit/backends/experimental/verl/gateway_host.py
```diff
@@ -0,0 +1,184 @@
...
+        import ray.util  # type: ignore[import-not-found]
+
+        ip = ray.util.get_node_ip_address()
+        # bracket-strip mirrors uni-agent's IPv6 handling
+        return ip.strip("[]")
+    except Exception:
+        return "127.0.0.1"
+
+
+def _serve_in_thread(app: web.Application, host: str, port: int, start_timeout: float = 15.0):
```
Linbo-Liu, line 109: As we discussed, this can be modularized and re-used. I will have a PR for it and we can switch it later.

src/agentcore_rl_toolkit/backends/experimental/verl/examples/math_agent/agentcore_agent.yaml
luyuzhe111: good instinct! the file is already built so that private values never belong in it. The two sensitive fields are env-var interpolations rather than literals:

```yaml
agent_runtime_arn: ${oc.env:AGENT_RUNTIME_ARN}
s3_bucket: ${oc.env:ACR_S3_BUCKET}
```

OmegaConf *errors* if the var is unset (rather than silently defaulting), so there's no incentive to hardcode a value to get a run going, and ARNs/buckets stay in the environment.

I kept it as a committed, runnable config rather than a `.example` because it doubles as the documented kwarg reference (`max_rollout_time`, `reward_mode`, `gateway_public_host`, ...), and the `.example` pattern would add a `cp` step before the example works. Happy to switch to `agentcore_agent.yaml.example` + gitignore if you'd rather have the structural guarantee. the trade-off is just that setup step.

src/agentcore_rl_toolkit/backends/experimental/verl/gateway_host.py
luyuzhe111: sg!

src/agentcore_rl_toolkit/backends/experimental/verl/agent_loop.py
luyuzhe111: yes exactly. this way we can directly specify

```
actor_rollout_ref.rollout.agent.agent_loop_config_path="$AGENT_LOOP_CONFIG"
```

to use agentcore without any changes to internal veRL

src/agentcore_rl_toolkit/backends/experimental/verl/gateway_host.py
```diff
@@ -0,0 +1,196 @@
...
+        super().log(request, response, time)
+
+
+@dataclass(frozen=True)
+class GatewayHandle:
+    """The per-process gateway singleton: the gateway, its verl-backed sampling
+    backend, and the base_url ACR agents should dial."""
+
+    gateway: RolloutGateway
+    backend: VerlSamplingBackend
```
Linbo-Liu, line 61: seems like `GatewayHandle.backend` is repeated with `GatewayHandle.gateway.backend`. Okay to keep it if we will refactor later.

luyuzhe111: yep it's the same object. the field exists to narrow the type: gateway.backend is the SamplingBackend protocol, which has no `pop_extra_fields` (that's VerlSamplingBackend-only, for verl's staleness tags), so dropping the field means a cast at the one call site instead. added a comment noting this. happy to revisit if your refactor changes how engine extras get out of the gateway!

Linbo-Liu (approved):

PR #92, opened 2026-07-27
Fix/slime sglang data parallel size

jackapbutler:
### Problem
SlimeRunner._build_slime_flags() doesn't pass --sglang-data-parallel-size to slime's train.py. Slime versions after commit fa3c990 added validation at sglang_utils/arguments.py:142 that reads args.sglang_data_parallel_size — an attribute that only exists if argparse registers it via the CLI flag. Since SlimeRunner never passed it, argparse never creates the attribute, and validation crashes with AttributeError.

Additionally, the [slime] extra was missing ray as a dependency despite SlimeRunner directly invoking ray start, ray stop, and ray job submit.

## Fix
   - Add sglang_data_parallel_size: int = 1 field to the SlimeRunner dataclass
   - Emit --sglang-data-parallel-size <value> unconditionally in _build_slime_flags()
   - Add ray to the [slime] optional dependency group in pyproject.toml

### Tests

   - Added two unit tests covering default and custom values for the new flag
   - Full test suite passes (173 tests)
Additionally this fix was tested for small scale RL experiments using a manual patch internally.

### DIFF
--- a/src/agentcore_rl_toolkit/backends/slime/runner.py
+++ b/src/agentcore_rl_toolkit/backends/slime/runner.py
@@ -87,6 +87,7 @@ class SlimeRunner:
     # Gateway use_sglang parsers (must match the served model) + cumulative mode.
     sglang_tool_call_parser: str = "qwen"
     sglang_reasoning_parser: str | None = None
+    sglang_data_parallel_size: int = 1
     cumulative_token_mode: bool = False
     renderer_family: str = "auto"
     gateway_log_level: str = "warning"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -63,6 +63,8 @@ slime = [
     # token IDs and logprobs into TraceRecords that slime consumes as Samples.
     # Required by agentcore_rl_toolkit.backends.slime.{gateway,rollout,traces}.
     "rllm-model-gateway>=0.1.0",
+    # SlimeRunner orchestrates training via ray (ray start, ray stop, ray job submit).
+    "ray",
```
lyzustc: I think this is not needed? In our installation guide https://awslabs.github.io/agentcore-rl-toolkit/guides/slime-backend-setup/#installation, choose either option A through slime official docker, or option B through install script https://github.com/awslabs/agentcore-rl-toolkit/blob/main/src/agentcore_rl_toolkit/backends/slime/scripts/install_slime.sh, we both need to install slime's dependencies, and ray is included (https://github.com/THUDM/slime/blob/main/requirements.txt#L18)

lyzustc (changes requested):
Thanks for contribution! Codes about adding sglang data parallel size look good to me, but I think there is no need to add ray to pyproject.toml as ray is already the dependency of slime package. Also please address merge conflicts in uv.lock.

jackapbutler:
Thanks @lyzustc, yes agreed on ray and have removed that commit and rebased on main

lyzustc (approved):

PR #94, opened 2026-07-28
feat(rollout-gateway): make renderer derender stages pluggable

Linbo-Liu:
*Issue #, if available:*

*Description of changes:*
`HfTemplateRenderer` derendered model output with one hardcoded path: split reasoning on `</think>`, then match tool calls with a regex for the `<tool_call><function=...>` XML format. Any model whose tool-call format differs silently produced zero tool calls. 

For example, Qwen3-0.6B has parsing format:
```
  <tool_call>
  {"name": <function-name>, "arguments": <args-json-object>}
  </tool_call>
```
while the parser can only handle
```
  <tool_call>\s*<function=NAME>...<parameter=K>V</parameter>...</function>\s*</tool_call>
```

  | Stage | Signature | Default |
  |---|---|---|
  | `reasoning_parser` | `raw_output -> (reasoning, body_text)` | split on `</think>` |
  | `tool_parser` | `(body_text, tools_schema) -> (text, tool_uses, ill_formed)` | `<tool_call><function=...>` regex |

The stages run in sequence — reasoning first, tool calls on what remains — matching the order the inference engines use. Overriding one leaves the other on its default, so a model needing engine-grade tool parsing but emitting plain `</think>` reasoning supplies only `tool_parser`. The tool stage is skipped entirely when a request carries no tools schema.

### DIFF
--- a/src/agentcore_rl_toolkit/rollout_gateway/render.py
+++ b/src/agentcore_rl_toolkit/rollout_gateway/render.py
@@ -9,8 +9,12 @@ inference backends (e.g. Tinker) that cannot render themselves.
 Two implementations behind one :class:`Renderer` protocol:
 
 * :class:`HfTemplateRenderer` (default, lightweight) — renders with the HF tokenizer's
-  ``apply_chat_template`` and parses output via ``parse_model_output`` (dependency-free
-  regex + ``</think>`` split). Needs only ``transformers``.
+  ``apply_chat_template`` and derenders by applying a reasoning parser then a tool
+  parser in sequence, each independently overridable via ``reasoning_parser`` /
+  ``tool_parser`` and each falling back to the dependency-free default (``</think>``
+  split / ``<tool_call><function=...>`` regex). Needs only ``transformers``; the
+  override seams are how engine-grade parsers (e.g. SGLang's detectors) plug in
+  without this package depending on any engine — the slime backend supplies them.
 * :class:`TinkerRenderer` — wraps a tinker-cookbook ``Renderer`` (install ``tinker``
   and ``tinker-cookbook`` manually; they require Python >=3.11). Its ``tinker_cookbook``
   import (which pulls torch) is deferred into ``__init__``, so importing this module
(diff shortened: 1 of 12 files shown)


tests/rollout_gateway/integration_test/sglang/test_e2e_qwen_sglang.py
```diff
@@ -105,19 +111,16 @@ def make_gateway():
...
-@asynccontextmanager
-async def serve(gateway: RolloutGateway):
-    from aiohttp import web
-
-    runner = web.AppRunner(gateway.app)
-    await runner.setup()
-    site = web.TCPSite(runner, "127.0.0.1", 0)
-    await site.start()
-    port = site._server.sockets[0].getsockname()[1]
+@contextmanager
```
luyuzhe111, line 114: wondering why we dropped the async here?

Linbo-Liu: We have a threaded server that can start rollout gateway even in a sync function call (both slime and verl invoke the gateway in sync caller), this integration test is to test whether threaded server can work in a sync caller.

src/agentcore_rl_toolkit/rollout_gateway/render.py
```diff
@@ -71,23 +78,38 @@ def parse(
...
         self,
         tokenizer,
         *,
         stop_sequences: list[str] | list[int] | None = None,
+        reasoning_parser: ReasoningParserFn | None = None,
+        tool_parser: ToolParserFn | None = None,
     ) -> None:
         self.tokenizer = tokenizer
         self._stop_sequences: list = list(stop_sequences) if stop_sequences else []
+        self.reasoning_parser: ReasoningParserFn = reasoning_parser or split_reasoning
```
lyzustc, line 111: Is it possible that we build a mapping from supported model family to parser functions?

Linbo-Liu: Do we want to implement the parser functions or import from inference engine? Our tenet is to have 0 dependency on vllm/sglang.

lyzustc: I am ok to use some lightweight packages like renderers (https://github.com/PrimeIntellect-ai/renderers), or implement and maintain ourselves. Anyway, I think it may improve user experience if we can automatically match the correct parser functions according to the model family used by users?

luyuzhe111 (approved):

PR #96, opened 2026-07-31
feat(slime): experimental slime backend on the rollout gateway

Linbo-Liu:
*Issue #, if available:*

*Description of changes:*
## Summary

Adds `backends/experimental/slime/` — a slime training backend built on the in-repo `rollout_gateway`. It plugs into **stock** `slime/train.py` through slime's public extension points only (`--rollout-function-path`, `--custom-reward-post-process-path`, `--custom-config-path`), so there is no forked trainer and no custom entrypoint. This is the successor to `backends/slime`, which stays untouched (it still depends on the external `rllm-model-gateway`) until this graduates.

## What's in it

**`integration/rollout.py`** — the custom rollout function (`generate_rollout`).
Serves one `RolloutGateway` in-process on a background thread (`ThreadedGatewayServer`), sampling token-in/token-out via SGLang's native `/generate` (`SglangHttpBackend` against slime's router) and rendering with the HF chat template of `--hf-checkpoint`. The gateway owns tokenization in both directions, so loss masking is well-defined and there's no retokenization drift between turns.

**`integration/rewards.py`** — `normalize_episode_rewards`, replacing slime's default reshape-based normalization, which assumes a fixed turn count per row. Aggregates rows to one reward per episode by `(group_index, gateway_session_id)`, normalizes across episodes within a task group, then writes the result back to every row — so a 3-turn success and a 2-turn failure each count as one data point. Strategies are pluggable (`grpo` default, `identity`); padding rows (`group_index=-1`) are skipped.

**`integration/sglang_parsing.py`** — derender seams built from SGLang's own detectors. Because the gateway samples via `/generate`, it bypasses the server's OpenAI layer, so reasoning/tool-call parsing has to happen gateway-side. `build_tool_parser` (`FunctionCallParser`) and `build_reasoning_parser` (`ReasoningParser`) are wired automatically from slime's existing `--sglang-tool-call-parser` / `--sglang-reasoning-parser` args, validated eagerly at gateway construction so a bad parser name fails at startup rather than on the first turn. Each stage is independent — unset one and the gateway keeps its dependency-free default. This lives in the slime backend, not `rollout_gateway`, because slime trainers always have sglang importable while the gateway package stays engine-free.

## Example-agent changes

Every `rl_app.py` now reads `api_key = payload["_rollout"].get("api_key", "EMPTY")` and passes it to its model client, so the gateway can key capture off the Bearer slot; `"EMPTY"` (the vLLM convention) keeps plain evaluation endpoints working unchanged. Also fixes response-text extraction in the officebench/appworld agents: `response.message["content"][0]["text"]` raises `KeyError` when the first content block is a tool-use or reasoning block, so text blocks are now filtered and joined.

`.gitignore`: broadened the local-config ignore to `backends/**/examples/**/config.yaml` (was slime-specific) and added `*.jsonl` for trace logs.

## Test plan

- Validated end to end against a live ACR deployment of `examples/strands_math_agent` — GSM8K GRPO with Qwen3-0.6B on 8×B200, colocated train+rollout, wandb-logged. Exact versions pinned in `SETUP.md` → Tested Versions.

Test accuracy increased from 6% -> ~80% after 100 GRPO steps.


src/agentcore_rl_toolkit/backends/experimental/slime/examples/math_agent/.wandb.env.example
```diff
@@ -0,0 +1,23 @@
+# Wandb configuration — copy to .wandb.env (gitignored) and fill in values.
+#
```
luyuzhe111, line 2: users should be able to run wandb cli login once and they won't need this API setup. can probably fold a note about wandb in the training script?

src/agentcore_rl_toolkit/backends/experimental/slime/examples/math_agent/train.sh
```diff
@@ -0,0 +1,171 @@
...
+# Usage:
+#   export SLIME_DIR=/root/slime \
+#          MODEL_DIR=/path/to/Qwen3-0.6B \
+#          TRAIN_DATA_PATH=/path/to/gsm8k_tiny.jsonl \
+#          MODEL_TYPE=qwen3-0.6B
+#   bash train.sh
+set -euo pipefail
+
+# === Paths (set these via env) ===
+SLIME_DIR="${SLIME_DIR:?Set SLIME_DIR (path to the slime repo, e.g. /root/slime)}"
```
luyuzhe111, line 18: hmm do we still need this

src/agentcore_rl_toolkit/backends/experimental/slime/examples/math_agent/.wandb.env.example
Linbo-Liu: SG

src/agentcore_rl_toolkit/backends/experimental/slime/examples/math_agent/train.sh
```diff
@@ -0,0 +1,171 @@
...
+MODEL_DIR="${MODEL_DIR:?Set MODEL_DIR (path to the HF model checkpoint)}"
+TRAIN_DATA_PATH="${TRAIN_DATA_PATH:?Set TRAIN_DATA_PATH (path to the training JSONL)}"
+VAL_DATA_PATH="${VAL_DATA_PATH:-${TRAIN_DATA_PATH}}"
+MODEL_TYPE="${MODEL_TYPE:?Set MODEL_TYPE (slime model-arch name, e.g. qwen3-0.6B)}"
+CONFIG="${CONFIG:-$(dirname "$0")/config.yaml}"
+
+# Set your cuda path. CUDA 13 only — cu12 is not supported.
+CUDA_HOME="${CUDA_HOME:-/usr/local/cuda-13.0}"
```
luyuzhe111, line 26: this should only impact compiler choice during installation? do we actually need this here?

src/agentcore_rl_toolkit/backends/experimental/slime/examples/math_agent/train.sh
```diff
@@ -0,0 +1,171 @@
...
+# time (the loader reads it once at startup), so a leftover Ray cluster would run
+# actors that never see it. Fail loudly now, not mid-training on "Multiple libcudart
+# libraries found".
+if pgrep -f 'raylet|gcs_server' >/dev/null 2>&1; then
+  echo "ERROR: Ray is still running after 'ray stop --force'." >&2
+  echo "  Kill it before rerunning: pkill -9 -f 'raylet|gcs_server|ray::'" >&2
+  exit 1
+fi
+
+# slime/ray/actor_group.py hardcodes the cu12-named torch_memory_saver preload .so,
```
luyuzhe111, line 53: wondering if this is still true with the latest slime version

src/agentcore_rl_toolkit/backends/experimental/slime/examples/math_agent/train.sh
Linbo-Liu: This is still needed. `pip install slime` will not install slime/train.py because it's outside its `src` dir: https://github.com/THUDM/slime/blob/main/train.py

src/agentcore_rl_toolkit/backends/experimental/slime/examples/math_agent/train.sh
```diff
@@ -0,0 +1,171 @@
...
+export CUDNN_PATH="${CUDNN_HOME}"
+
+export PYTHONUNBUFFERED=1
+ray start --head --num-gpus ${NUM_GPUS} --disable-usage-stats
+
+# Source model architecture args (populates MODEL_ARGS)
+source ${SLIME_DIR}/scripts/models/${MODEL_TYPE}.sh
+
+# === Launch training ===
+export no_proxy=127.0.0.1
```
luyuzhe111: wondering what does this do

src/agentcore_rl_toolkit/backends/experimental/slime/integration/rewards.py
```diff
@@ -0,0 +1,102 @@
...
+
+Usage:
+    --custom-reward-post-process-path \
+        agentcore_rl_toolkit.backends.experimental.slime.integration.rewards.normalize_episode_rewards
+"""
+
+from collections import defaultdict
+
+
+def _grpo_normalize(episode_rewards: list[float], std_normalize: bool) -> list[float]:
```
luyuzhe111: i'm a bit surprised we actually need to implement this? slime should support black-box agent with reward coming from the session already? can we fold the reward extraction in the rollout function so that we can just reuse reward processing logics from slime?

src/agentcore_rl_toolkit/backends/experimental/slime/integration/rollout.py
```diff
@@ -0,0 +1,639 @@
...
+            ),
+        )
+
+
+# ---------------------------------------------------------------------------
+# Lazy imports
+# ---------------------------------------------------------------------------
+
+
+def _import_slime_types():
```
luyuzhe111: this try except doesn't make much sense to me.

src/agentcore_rl_toolkit/backends/experimental/slime/integration/rollout.py
```diff
@@ -0,0 +1,639 @@
...
+    mechanics live in the shared :class:`ThreadedGatewayServer`.
+
+    ``tool_call_parser`` / ``reasoning_parser`` (slime's --sglang-tool-call-parser /
+    --sglang-reasoning-parser) select SGLang's engine-grade parsers for the served
+    model's output format; the gateway samples via the native /generate endpoint,
+    so that parsing happens here rather than in the SGLang server. Each is
+    independent: an unset one leaves that derender stage on the gateway's
+    dependency-free default.
+    """
+    try:
```
luyuzhe111: try except also meaningless here?

src/agentcore_rl_toolkit/backends/experimental/slime/integration/rollout.py
```diff
@@ -0,0 +1,639 @@
...
+        return rewards[-1] if rewards else 0.0
+    return float(rewards)
+
+
+# ---------------------------------------------------------------------------
+# TraceRecord -> slime Sample conversion
+# ---------------------------------------------------------------------------
+
+
+def _make_noop_sample(group_index: int = -1, session_id: str = "", status_name: str = "COMPLETED"):
```
luyuzhe111: do we still need this

src/agentcore_rl_toolkit/backends/experimental/slime/scripts/install_slime.sh
```diff
@@ -0,0 +1,101 @@
...
+: "${CUDA_HOME:=/usr/local/cuda-13.0}"
+export CUDA_HOME
+
+# The source builds below (flash-attn, transformer-engine, apex) write tens of GB
+# of nvcc scratch files to $TMPDIR and can fill up a small root volume
+# ("No space left on device" from cc1plus/nvcc). Prefer the instance's large
+# ephemeral NVMe volume (present on AWS DLAMI/HyperPod nodes) unless the caller
+# already set TMPDIR.
+if [ -z "${TMPDIR:-}" ] && [ -d /opt/dlami/nvme ]; then
+  TMPDIR="/opt/dlami/nvme/${USER}/tmp"
```
luyuzhe111: TMPDIR remapping should be in .bashrc

src/agentcore_rl_toolkit/backends/experimental/slime/scripts/install_slime.sh
```diff
@@ -0,0 +1,101 @@
...
+uv pip install -e . --no-build-isolation --config-settings editable_mode=compat
+cd ..
+
+uv pip install --reinstall-package nvidia-cutlass-dsl-libs-base --no-deps \
+  "nvidia-cutlass-dsl-libs-base==4.5.2"
+
+uv pip install --reinstall-package pyjwt PyJWT
+
+# Install slime
+git clone https://github.com/THUDM/slime.git
```
luyuzhe111, line 63: we actually need a local slime copy? :))))

src/agentcore_rl_toolkit/backends/experimental/slime/examples/math_agent/train.sh
Linbo-Liu: Yes. https://github.com/THUDM/slime/blob/main/slime/ray/actor_group.py#L64-L84

src/agentcore_rl_toolkit/backends/experimental/slime/scripts/install_slime.sh
```diff
@@ -0,0 +1,101 @@
...
+# Apply slime's official patches to megatron + sglang.
+SLIME_PATCH_DIR="$(cd slime/docker/patch/latest && pwd)"
+SITE_PACKAGES="$(python -c 'import sysconfig; print(sysconfig.get_path("purelib"))')"
+
+# Megatron patch: apply against the repo without leaving cwd (git -C).
+git -C Megatron-LM update-index --refresh >/dev/null 2>&1 || true
+git -C Megatron-LM apply --3way "$SLIME_PATCH_DIR/megatron.patch"
+
+# sglang patches: apply into site-packages without leaving cwd (patch -d).
+patch -d "$SITE_PACKAGES" -p2 -F0 -N < "$SLIME_PATCH_DIR/sglang.patch"
```
luyuzhe111, line 85: what patch do we need exactly?

PR #104, opened 2026-08-10
fix(verl): preserve full-context agent trajectories

luyuzhe111:
## Problem

The verl integration treated `response_length` as both the per-turn generation cap and the cumulative trajectory budget. Together with verl's fixed prompt and response tensor regions, this forced a static context partition: short prompts wasted available context, while prompts that exceeded `prompt_length` could be truncated during output conversion.

## Fix

- Separate the model window (`max_model_len`), cumulative trajectory storage (`response_length`), and per-turn generation cap (`max_tokens_per_turn`).
- Preserve prompt overflow at the start of verl's response region with zero loss mask and zero rollout logprob, retaining token order without training on input tokens.
- Validate token-budget configuration and `TraceRecord` alignment at their ownership boundaries.
- Document the full-window configuration and make `max_tokens_per_turn` explicit in the math-agent example.

### Example

Given `prompt_length=4`, `response_length=max_model_len=12`, and two model turns separated by a tool result, the gateway may capture:

```text
initial prompt:  [SYS, RULES, USER, QUESTION, DETAILS, ASK]
response tokens: [CALL, CALC, TOOL_42, ANSWER, 42]
loss mask:       [   1,    1,       0,      1,  1]
```

Previously, verl's four-token prompt region retained only `[USER, QUESTION, DETAILS, ASK]`, dropping `SYS` and `RULES`. The fix preserves the original order by moving the overflow into the response prefix:

```text
prompt_ids:        [SYS, RULES, USER, QUESTION]
response_ids:      [DETAILS, ASK, CALL, CALC, TOOL_42, ANSWER, 42]
response_mask:     [      0,   0,    1,    1,       0,      1,  1]
response_logprobs: [      0,   0,   lp,   lp,       0,     lp, lp]
```

The overflow remains visible as model input but contributes no policy loss or rollout logprob. Tool output remains zero-masked, while both assistant turns remain trainable.

## Validation

- Added focused tests for prompt-overflow preservation, zero masking/logprobs, gateway context budgeting, invalid budgets, and malformed trace records.
- Completed a 116-step Qwen3-4B GSM8K run on 8 B200 GPUs with `prompt_length=14336`, `response_length=max_model_len=16384`, and `max_tokens_per_turn=2048`.
- Against the prior `14336` prompt / `2048` response configuration, validation accuracy is roughly the same.

Screenshot 2026-08-12 at 9 16 56 PM

## Follow-up

- **Context overflow:** Return an explicit error to the agent when a newly rendered prompt fills or exceeds the remaining model context, while retaining the tool result in the captured conversation and excluding the unsampled turn from training.
- **Higher padding buffer width:** To avoid confusing `prompt_length/clip_ratio` metrics, we encourage users to use a large `prompt_length` close to `max_model_len`. This causes verl to over-provision padding width, though the padding will be removed before model forward pass to avoid computation waste. Follow-ups can focus on addressing this limitation.

### DIFF
--- a/src/agentcore_rl_toolkit/backends/experimental/verl/agent_loop.py
+++ b/src/agentcore_rl_toolkit/backends/experimental/verl/agent_loop.py
@@ -106,6 +106,7 @@ class AgentCoreAgentLoop(AgentLoopBase):
         *,
         agent_runtime_arn: str,
         s3_bucket: str,
+        max_tokens_per_turn: int,
         exp_id: str | None = None,
         tps_limit: int = 5,
         max_pool_connections: int = 100,
@@ -141,6 +142,12 @@ class AgentCoreAgentLoop(AgentLoopBase):
 
         self.prompt_length = self.rollout_config.prompt_length
         self.response_length = self.rollout_config.response_length
+        self.max_model_len = self.rollout_config.max_model_len
+        self._validate_token_budgets(max_tokens_per_turn)
+        self.max_tokens_per_turn = max_tokens_per_turn
+        # The gateway cannot emit more tokens than verl's response region can store.
+        # The two padded regions may sum past max_model_len; valid tokens never do.
+        self.max_context_tokens = self.response_length
         self.max_rollout_time = max_rollout_time
         # Kept as config, not inlined, so a validated trainer-side mode can land
         # without a signature change. Reward semantics: see the README.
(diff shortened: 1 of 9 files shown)


src/agentcore_rl_toolkit/backends/experimental/verl/agent_loop.py
```diff
@@ -171,6 +178,35 @@ def __init__(
...
+                f"rollout.max_model_len ({self.max_model_len})"
+            )
+        if self.response_length > self.max_model_len:
+            raise ValueError(
+                f"rollout.response_length ({self.response_length}) cannot exceed "
+                f"rollout.max_model_len ({self.max_model_len})"
+            )
+        if (
+            not isinstance(max_tokens_per_turn, int)
+            or isinstance(max_tokens_per_turn, bool)
```
hellodanylo: why bool?

src/agentcore_rl_toolkit/backends/experimental/verl/agent_loop.py
```diff
@@ -338,19 +374,22 @@ def _record_to_output(
...
-        response_ids = response_ids[: self.response_length]
-        response_mask = response_mask[: self.response_length]
-        response_logprobs = response_logprobs[: self.response_length]
+
+        # verl stores prompts and responses in fixed-width regions. Preserve the
+        # complete token order when the initial prompt exceeds its region by placing
+        # the overflow at the start of the response region. These are input tokens:
+        # they remain attended to, but carry no policy loss or rollout logprob.
+        prompt_ids = initial_prompt[: self.prompt_length]
+        prompt_overflow = initial_prompt[self.prompt_length :]
```
hellodanylo, line 385: This could make the Verl metrics incorrect though? Both prompt and response length are logged to mlflow/wandb. From the customer's perspective, I would recommend to just set prompt length high enough that it never needs overflow.

hellodanylo (approved):

src/agentcore_rl_toolkit/backends/experimental/verl/agent_loop.py
luyuzhe111: ah it's for a corner case where boolean value might be recognized as 0 / 1 integers since boolean subclass int. I collapsed this into an exact type check for clarity!

src/agentcore_rl_toolkit/backends/experimental/verl/agent_loop.py
luyuzhe111: yea agreed. also flagged this metric semantic change as one of the follow-ups in the PR. I can set prompt length to max model length. the only downside is probably just double the padding buffer. the overflow mechanism can still be a good fallback option. does  this sound good to you?

luyuzhe111: have added a note in verl integration README to recommend a high prompt length value!

lyzustc (approved):

PR #118, opened 2026-08-30
chore(verl): upgrade to verl 0.9.0 and vLLM 0.24.0

luyuzhe111:
## Summary

- Replace the git-pinned verl snapshot with the `verl==0.9.0` release and upgrade vLLM from 0.23.0 to 0.24.0, which includes the fix from [vLLM #44483](https://github.com/vllm-project/vllm/pull/44483).
- Keep `transformers==5.12.1` for the rollout gateway parser while scoping the dependency override to verl instead of overriding the environment globally.
- Simplify the example launch configs by removing defaults and legacy flags, and select the Triton MoE backend required by the migration-agent setup.

## Validation

- MigrationBench Megatron LoRA E2E completed 30 training steps with no rollout failures, aborted responses, OOMs, NCCL errors, or tracebacks. Checkpoints and validation completed successfully at steps 15 and 30.
- Validation reward remained comparable to the previous stack:

| Step | Previous | Upgraded | Delta |
| ---: | ---: | ---: | ---: |
| 0 | 0.4130 | 0.4264 | +1.34pp |
| 15 | 0.5786 | 0.5619 | -1.67pp |
| 30 | 0.6756 | 0.6522 | -2.34pp |

Both runs show the same improving trend, with the upgraded stack within 2.4 percentage points through step 30.


src/agentcore_rl_toolkit/backends/verl/examples/migration_agent/megatron_lora_sync_grpo.sh
```diff
@@ -65,9 +59,9 @@ PROJECT_NAME=${PROJECT_NAME:-agentcore_grpo}
 EXPERIMENT_NAME=${EXPERIMENT_NAME:-migrationbench_qwen3_coder_30b}
 CKPTS_DIR=${CKPTS_DIR:-checkpoints/${PROJECT_NAME}/${EXPERIMENT_NAME}}
 
+# moe_backend=triton is required to avoid the default FlashInfer MoE kernel failure.
```
hellodanylo: I think FlashIner MoE is only a problem for Blackwell. It works on Hopper for me.

hellodanylo (approved):

src/agentcore_rl_toolkit/backends/verl/examples/migration_agent/megatron_lora_sync_grpo.sh
luyuzhe111: Good point. Scoped the comment to the AWS P6-B200 (NVIDIA Blackwell) setup where we observed the FlashInfer MoE failure.

luyuzhe111: kk added a qualifier to the comment on b200. thanks!

PR #120, opened 2026-09-02
docs: propose a runtime invocation protocol

luyuzhe111:
### Summary

This PR adds a roadmap and design proposal for a Runtime Invocation Protocol: a shared execution lifecycle for stateful, long-running workloads on AgentCore Runtime. The protocol provides the common Runtime capability needed by both the Rollout SDK and the Sandbox SDK.

The proposal separates three identities that are currently easy to conflate:

- `runtimeSessionId` identifies the sticky compute and isolation scope;
- `conversation_id` optionally correlates application-managed conversation history; and
- `invocation_id` identifies one execution with its own lifecycle.

## Proposed behavior

- Treat foreground and background as delivery modes over the same invocation. Foreground waits on the initial connection; background returns a handle for later polling.
- Give every invocation explicit status, retry, interruption, and terminal-result semantics.
- Generate the invocation ID before submission and reuse it for retries. Repeated submission addresses the same execution for as long as its invocation record survives.
- Keep stopping local waiting and stopping the Runtime session as separate operations.
- Persist invocation state for both delivery modes through one filesystem store. A microVM-local root is the baseline implementation; placing the same file layout on managed session storage extends recovery across compute replacement. S3 remains available for direct external retrieval, larger artifacts, or independent retention.
- Allow multiple invocations within one Runtime session without making the protocol responsible for application-level ordering or shared-state concurrency.

The design does not claim exactly-once external side effects, recovery after the selected store is lost, conversation-history management, or replay of a disconnected foreground stream.

## SDK impact

For the Rollout SDK, this preserves long-running background rollouts while adding foreground invocation and sticky-session follow-up. Generic task lifecycle, persistence, and retry move below the rollout-specific layer, leaving the Rollout SDK responsible for rollout configuration, batching, grouping, result policy, and training integration.

For the Sandbox SDK, the same lifecycle enables both foreground and detached `exec()` calls with invocation handles, polling, and recovery within the selected store's lifetime. Sandbox provisioning, shell and file UX, and native interactive shells remain separate concerns.

The proposal uses OpenAI Responses as a reference pattern for separating execution identity from connection mode, and compares E2B and AgentCore Interactive Shell as related process and connection models. It does not attempt to clone any of those APIs.

## Scope

This PR is documentation only. Generic cancellation is explicitly outside the initial protocol because application handlers and their child resources cannot be reliably stopped through one cross-workload contract. Consumer persistence defaults, public class names, the Sandbox process-manager transport, filesystem retrieval mechanics, and result-size boundaries remain open implementation questions. The intended long-term home for the generic capability is the official AgentCore SDK or service; ART provides the initial design and concrete Rollout and Sandbox use cases.


roadmaps/draft/runtime_invocation_protocol.md
```diff
@@ -0,0 +1,1061 @@
...
+invocation and accept follow-up work. A `runtimeSessionId` therefore cannot
+also serve as the identity and lifecycle of every execution inside the session.
+
+This proposal defines an explicit invocation lifecycle nested within a Runtime
+session. Every invocation has its own ID, status, terminal result, retry
+semantics, and cancellation target. Foreground and background are two ways to
+interact with that same lifecycle:
+
+- `background=false` starts the invocation and waits on the initial
+  connection; and
```
hellodanylo, line 24: Do we expect the server to be parallel? For example, I send one background=false invocation, and while it's still in-progress, send another one. Can I expect the second invocation to make any progress before the first one finishes?

roadmaps/draft/runtime_invocation_protocol.md
```diff
@@ -0,0 +1,1061 @@
...
+  protecting protocol state from arbitrary code inside the same Runtime
+  session.
+- Replaying foreground stream contents after a disconnect in the initial
+  implementation.
+
+## Identity model
+
+RIP distinguishes three identities:
+
+- **Runtime session ID**: the AgentCore compute, sticky-routing, isolation,
```
hellodanylo, line 237: It will be great to have an implementation that works against Docker containers for local testing (including CI).

roadmaps/draft/runtime_invocation_protocol.md
```diff
@@ -0,0 +1,1061 @@
...
+# Illustrative shape, not a final class name.
+result = client.invoke(request, background=False)
+
+handle = client.invoke(request, background=True)
+result = handle.result(timeout=600)
+```
+
+An asyncio client can expose the same remote modes:
+
+```python
```
hellodanylo, line 295: It would be more idiomatic to have two different Python classes, one sync, one async. Same method names, so we can avoid _async suffix.

roadmaps/draft/runtime_invocation_protocol.md
```diff
@@ -0,0 +1,1061 @@
...
+Every invocation exposes at least:
+
+```text
+absent
+  |
+  | start
+  v
+in_progress ------> completed
+      |
+      +-----------> failed
```
hellodanylo: What does failed here mean? Does it have more semantics besides adding a second result type (besides completed)? If not, can this be wrapped into a binary Result type in completed state?

roadmaps/draft/runtime_invocation_protocol.md
```diff
@@ -0,0 +1,1061 @@
...
+may supply or persist one when they need to transfer a handle between
+processes, retry after losing local state, or reattach from another client.
+
+### Waiting, cancellation, and session lifecycle
+
+RIP keeps three actions distinct:
+
+- **Stop waiting**: a local timeout returns control to the caller. The remote
+  execution continues by default.
+- **Cancel the invocation**: the execution adapter cooperatively stops the
```
hellodanylo: What are the semantics for the cancellation being complete? What if it takes long time to cancel? Would it race with the next invocation?

roadmaps/draft/runtime_invocation_protocol.md
```diff
@@ -0,0 +1,1061 @@
...
+them as distinct executions. RIP does not serialize requests or isolate
+application state: preventing concurrent invocations from corrupting shared
+files, in-memory objects, or conversation history remains the application's
+responsibility.
+
+### Streaming
+
+Stream recovery is future work. The first implementation may preserve native
+foreground streaming while the connection remains open, but RIP recovery
+guarantees only lifecycle status and a bounded terminal result. It does not
```
hellodanylo: A stream is not well defined here. Without additional assumptions, HTTP responses are all or nothing, and a disconnect would not return anything that a client could really use, unless we are talking about server-side events or some other streaming protocol. The spec needs to be a little more precise about what the streaming output is.

roadmaps/draft/runtime_invocation_protocol.md
```diff
@@ -0,0 +1,1061 @@
...
+reads, larger artifacts, or retention beyond the managed session-storage
+lifecycle.
+
+The public handle should expose neither the storage backend nor the retrieval
+transport. Keeping those choices internal lets adapters evolve without
+changing invocation identity or workload APIs.
+
+### Recovery
+
+A client can retrieve a terminal result after the initial connection closes or
```
hellodanylo: How does a client receive a response with the handle ID but remains polling for its result? The exact HTTP transport is not clear yet.

roadmaps/draft/runtime_invocation_protocol.md
```diff
@@ -0,0 +1,1061 @@
...
+  failure, and storage retention.
+- Define artifact policies for results that exceed the manifest format.
+- Evaluate moving the generic app, client, and protocol behavior into
+  `bedrock-agentcore-sdk-python`.
+- Keep Rollout, Sandbox, Rollout Gateway, and trainer-specific integrations in
+  ART.
+
+## Open questions
+
+- Should the Sandbox client use `InvokeAgentRuntime` or
```
hellodanylo, line 773: Process launch latency could create too much overhead, especially in python, during polling.

roadmaps/draft/runtime_invocation_protocol.md
```diff
@@ -0,0 +1,1061 @@
...
+
+The client initially performs this operation through another
+`InvokeAgentRuntime` call using the same Runtime session ID. This path must be
+validated after the original execution environment has become idle.
+
+### App-handler `cancel`
+
+For a live asynchronous handler, the adapter can retain its task, set a
+cooperative cancellation signal, and request task cancellation. A synchronous
+handler running in a Python worker thread cannot be safely force-stopped.
```
hellodanylo: Async Python tasks can not be force-cancelled either. An async task can choose to ignore the cancellation exception, including due to a bug. An async task can also take an unlimited amount of time to stop. Further, even if a task did finish , it does not guarantee that other related resources have stopped -- subprocesses, other tasks, etc. In general, I would recommend against making "cancel" operation, because its semantics are nearly impossible to specify and enforce in general without significant complexity. The only real cleanup path is a full VM restart, which is already available with AgentCore API. 

I also can't quite think of a use case for cancelling invocation either. The clients of RIP SDK are free to have their own cleanup functions, however. For example, they can pkill subprocesses or send "stop" events to async agent harnesses, etc. I just don't see how RIP protocol can promise anything both generic and useful at the same time.

roadmaps/draft/runtime_invocation_protocol.md
```diff
@@ -0,0 +1,1061 @@
...
+semantics, and cancellation target. Foreground and background are two ways to
+interact with that same lifecycle:
+
+- `background=false` starts the invocation and waits on the initial
+  connection; and
+- `background=true` starts the invocation and returns a handle before it
+  finishes.
+
+Invocation state and bounded terminal results must survive the loss of
+process-local state in both modes. Managed session storage is the proposed
```
hellodanylo: Do we have a concrete use case for surviving VM restarts? It would be great to have it documented. Neither Rollout SDK, nor Sandbox SDK seem to need it directly.

hellodanylo (changes requested):
Great doc!
Minor things to address:
* it would be great to document a use case for persisted state across VM restarts, so we can validate the design against it
* failed and cancelled states seem to be not well defined, I recommend to drop or provide addition specification

hellodanylo (approved):
Good to commit the doc as is despite the comments.

roadmaps/draft/runtime_invocation_protocol.md
luyuzhe111: good question. the `BedrockAgentCoreApp` does provide a thread pool so it should work? i have not tested it tho.

roadmaps/draft/runtime_invocation_protocol.md
luyuzhe111: ah good point. will consider it in the actual implementation!

roadmaps/draft/runtime_invocation_protocol.md
luyuzhe111: make sense will update the example code.

roadmaps/draft/runtime_invocation_protocol.md
luyuzhe111: good point. just folded it to "completed".

roadmaps/draft/runtime_invocation_protocol.md
luyuzhe111: good question. i agree with your high-level comments that the cancellation operation is a bit hard to implement & semantics are also currently not well defined. so i have removed the discussions about cancellation in this doc.

roadmaps/draft/runtime_invocation_protocol.md
luyuzhe111: good point. I was using “streaming” too loosely here. I’ve clarified that RIP v1 does not define an incremental streaming protocol. An adapter may preserve an existing SSE or EventStream response while the connection remains open, but after a disconnect RIP only supports retrieving the invocation status and terminal result. it does not replay the original response stream.

roadmaps/draft/runtime_invocation_protocol.md
luyuzhe111: not sure i followed your question actually. could you clarify a bit

roadmaps/draft/runtime_invocation_protocol.md
luyuzhe111: for `InvokeAgentRuntime`, I'm actually planning to extend the agentcode-sandboxd, which is written in Go. It's currently only a health shim but could do more than that?

roadmaps/draft/runtime_invocation_protocol.md
luyuzhe111: thanks for the advice & explanation! i have removed the `cancel` operation in the proposal, as previously noted.

roadmaps/draft/runtime_invocation_protocol.md
luyuzhe111: i was mostly targeting cases where the session idles out due to delayed polling (say we have very high concurrency). it's also possible that the server used to satisfy the `/ping` contract gets killed by the agent, where the session will be considered unhealthy and gets lost. this should apply to both rollout and sandbox sdk.  though i guess we can still infer that some sort of crash happened if polling of an invocation that should have started doesn't exist?

either way, i think it's indeed much easier to start with in-session disk storage. and if we truly need managed session storage that survives between restarts in the future, we can actually easily enabled that by changing the save dir to the mounted persistent storage (from `/tmp/.agentcore/started.json` to `/mnt/workspace/.agentcore/started.json`)?

have updated the doc correspondingly!

roadmaps/draft/runtime_invocation_protocol.md
hellodanylo: Right, I forgot it's not Python. All good!

PR #121, opened 2026-09-04
chore: standardize agent skill location

luyuzhe111:
## Summary

- Move the canonical CloudWatch session log skill from `.claude/skills` to the standard `.agents/skills` location.
- Keep Claude compatibility through a relative symlink.
- Preserve the original skill content unchanged.

## Validation

- Pre-commit hooks passed.
- Commitizen validation passed.
- Verified the canonical skill content is byte-for-byte identical to the original file.
- Verified the compatibility symlink resolves correctly.
- `git diff --check` passed.


hellodanylo (approved):
LGTM

PR #122, opened 2026-09-04
fix(verl): drop trace-less failed rollouts

luyuzhe111:
## Motivation

When an AgentCore rollout failed before producing any `TraceRecord`, the agent loop created a one-token pad trajectory with `response_mask=[1]`. Although its reward and log probability were zero, the token still entered actor loss and GRPO group statistics, producing invalid advantage, rollout-correction, and importance-ratio signals.

## Summary

- Raise when a rollout produces no real trace instead of storing a synthetic training row.
- Preserve the existing partial-failure behavior: if real trace data was captured before failure, train it with reward zero.
- Allow GRPO to operate on the real rows that remain. If a training batch has no materializable trajectory, allow the existing failure to surface.

## verl failed-group behavior

The V1 TQ worker marks the prompt group as failed if any session raises, after allowing all sibling sessions to settle. This PR does not change verl's existing group-level failure policies:

- Synchronous replay with `sync_refill_failed_groups=false` (the default and current configuration) keeps failed groups sampleable. Any successful sibling rows are materialized and GRPO uses only those remaining real rows.
- Synchronous replay with `sync_refill_failed_groups=true` refills failed groups only when they contain no materializable trajectory. Partial groups with successful sibling rows remain usable.
- Asynchronous replay evicts and refills every failed prompt group, including any successful sibling rows. Supporting partial failed groups in async training would require a separate change to verl's replay-buffer policy.

This implementation intentionally prioritizes training correctness over rollout efficiency. In asynchronous training, that can discard valid sibling work and require a refill, but it avoids admitting a fabricated row into actor loss or GRPO statistics. A synthetic row sharing the prompt UID affects GRPO mean and standard deviation even if its response mask is zero, because group statistics are computed before masking. verl's safe batch padding instead uses an independent padding UID downstream, after trajectory materialization; using that mechanism to represent a failed session would require coordinated TQ worker and replay-buffer changes.

## Testing

- `pytest tests/backends/verl/` — 53 passed.
- `pre-commit run --all-files` — passed.
- Real AgentCore math-agent smoke test covered both trace-less failure and successful trace capture.
- Synchronous verl V1 mixed-session harness retained two successful rows after one session failed and computed GRPO advantages from the remaining rows.

### DIFF
--- a/src/agentcore_rl_toolkit/backends/verl/README.md
+++ b/src/agentcore_rl_toolkit/backends/verl/README.md
@@ -37,11 +37,6 @@ verl main_ppo (v1) ──> AgentLoopWorker ──> AgentCoreAgentLoop.run()
 - A session's trajectory tree can fork (sub-agents, context compaction); every leaf
   becomes its own training row (`run()` returns `list[AgentLoopOutput]`) — hence the
   hard `trainer.use_v1=true` requirement.
-- Failed rollouts (timeout, ACR error, agent error) never raise. They yield a
-  single-pad-token fallback row: `response_mask=[1]` (verl's rollout-correction
-  helper requires ≥1 valid response token per row — an all-zero mask is rejected),
-  logprob 0, reward 0. See the TODO below for its current gradient semantics.
-
 The agent must forward the trainer-supplied key when it constructs its model client:
 
 ```python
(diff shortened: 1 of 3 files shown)


hellodanylo (approved):
LGTM, but we will need to resolve the async case in the future.

PR #129, opened 2026-09-09
Feat/reward extra info

xandernewton:
Populates reward_extra_info from container session which is currently left as empty.

```
   for k in self._rei_keys_seen:
            info.setdefault(k, 0.0)
        self._rei_keys_seen.update(info)
```
The keys for each session are set to zeros and then updated with the actual key values from each session. If a session does not return all the keys (possibly due to error in rollout) verl fills the missing value with None and then will crash when computing mean across values. Therefore setting missing values to zero ensures this does not occur.


lyzustc (approved):

hellodanylo (approved):
LGTM

PR #130, opened 2026-09-09
feat: SWE Agent Training Recipe with Async Verl Trainer

hellodanylo:
# feat: SWE Agent Training Recipe with Async Verl Trainer

## Summary

PR #123 landed the agent side of the OpenHands SWE agent: the container image, the HTTP agent server, the `RolloutSession` lifecycle abstraction (`agentcore` and `docker` backends), the shared concurrency primitives, and a local batch-evaluation driver (`examples/openhands_swe_agent/rollout_batch.py`). This PR adds the verl side: a new `RolloutSessionAgentLoop` that drives a `RolloutSession` from inside verl's v1 agent-loop worker, the Ray-hosted shared resources that bound it cluster-wide, two trainer metric mixins, and a complete Verl configuration set for GRPO training of Qwen3-Coder-30B on SWE-Gym.

### `backends/experimental/verl/` — the loop and its resources

`RolloutSessionAgentLoop` (registered as `rollout_session_agent_loop`) holds only the verl-facing contracts and delegates the container lifecycle to a composed `RolloutSession` selected by `rollout_session_backend.backend`. Per rollout it opens a gateway session keyed by its own `session_id`, builds the task dict handed to the container (including the LiteLLM-shaped `llm` block pointing at the local gateway, with `api_key=session_id` as the capture key), runs `run_rollout_with_bounds`, then drains the gateway into `TraceRecord`s and reshapes them into `AgentLoopOutput`s.

`rollout_session_resources.py` holds both halves of one contract in one file: the driver creates the run's shared `LocalPrioritySemaphore` / `LocalPriorityAssigner` / `ACRRateLimiter` as *named* Ray actors, and every agent-loop worker looks them up by the same names via `get_rollout_session_bounds`. Process-local counters would not cap anything across workers; named actors give one counter per run.

### `backends/experimental/verl/trainer_mixins/` — two independent metric features

`AgentLoopMetricsMixin` fetches `extra_fields` for the step's live TransferQueue entries and reduces every entry of each sample's `metrics` dict to `agent_loop/<name>/{mean,min,max,sum}`. Names are not hardcoded, so any metric the agent harness leaves behind flows through unchanged. It collapses to one representative row per `request_id` first, otherwise a session with many trajectories would outweigh a session with one; and it reduces each metric only over the sessions that actually reported it, so an absent metric is excluded rather than counted as 0.0.

`AdvantageZeroMetricsMixin` adds `critic/advantages/zero_mean` — the fraction of valid (response-masked) advantage entries that are exactly zero, i.e. how much of the step's loss signal carries no gradient — and `critic/advantages/zero_pass_mean`, which says which way the collapsed GRPO groups went (all-pass vs all-fail). The second one is the diagnostic that matters on SWE tasks, where a group collapsing because every rollout failed and a group collapsing because every rollout passed call for opposite responses.

### `backends/verl/examples/swe_agent/` — the recipe

`main.py` is the entrypoint: calls `run_ppo` with the custom task runner. `trainer.py` registers `art_sync` / `art_colocate_async` / `art_separate_async` — verl's three v1 backends with the two mixins layered on.

The config is composed from single-concern groups so a run is a defaults list rather than a wall of overrides: cluster shape (`cluster_trainer_rollout_n{1,2,4}`, `cluster_trainer2_rollout1`), dataset (`dataset_swe_gym`), model/parallelism (`qwen3_30b_p5`, `qwen3_30b_p6`), engine (`megatron`, `megatron_offload`, `megatron_recompute`, `lora`), algorithm (`smart_ppo`, `rollout_correction_decoupled`), async mode (`async_separate_b2`), and the loop itself (`rollout_session_agent_loop`). The shipped `main.yaml` is GRPO + Megatron + LoRA + separate-async on SWE-Gym at 64k context, `n=16`.

## Design decisions worth calling out

**One trainable record per rollout.** verl's v1 trainer does not yet handle the multi-record case well, so the recipe configures the gateway with `history_mode: linear` and `fork_threshold_tokens: 0` to linearize the trajectory tree, and the loop keeps the record with the most trainable tokens (`max` by loss-mask sum). `num_records` is still recorded in the session meta so a run that unexpectedly produces more than one is visible rather than silent.

**A failed rollout still emits a sample.** Container-reported failures (the container's own stack trace is already in the S3 dump, so it is recorded and aborted rather than re-raised) and trainer-side failures both produce a placeholder `AgentLoopOutput` with `reward_score=0`, which keeps the GRPO group intact. Its `min/max_global_steps` are stood in by the dispatch step rather than the current `global_steps`, so `trajectory_staleness` stays comparable to real data instead of being inflated by rows that never had a weight version.

**Prompt/response split derives from the loss mask.** `len(record.loss_mask)` is the response length; everything before it is prompt. Because `response_length` is the whole-trajectory budget, the per-record response budget is `response_length - len(prompt_ids)`, and a prompt that fills the entire budget yields an empty response — which crashes verl's `AgentLoopOutput.as_dict`, so the loop emits one masked token instead.

**Two observability sinks per rollout.** A `PersistentDict` backed by DynamoDB carries live per-session state (priorities, slot start times, span timings, timeout flags, EC2 instance), and the full `RolloutOutput` — task, loop outputs, exception, meta, container dump — goes to S3 under `{experiment_start_at}/{session_id}`. Both are optional: no `dynamodb_table` falls back to a `NullPersister`.

## Known gaps

- No tests accompany the new loop, resources, or mixins. The loop is hard to test without verl, but `make_successful_loop_output` / `make_failed_loop_output` and the two mixins' reduce logic are pure enough to cover.

## Testing

* When training for 70 steps, SWE Gym pass rate of Qwen3 Coder 30B improves from 55% to 69%. A detailed report will be released later.
* `tests/rollout_session/rollout_session_bounds_test.py` passes unchanged apart from the rename. The recipe itself has been exercised by running it — no automated coverage of the new verl-side code yet (see above).


src/agentcore_rl_toolkit/backends/experimental/verl/trainer_mixins/advantage_metrics.py
```diff
@@ -0,0 +1,109 @@
...
+import logging
+import os
+from typing import TYPE_CHECKING, Any
+
+import torch
+import transfer_queue as tq
+
+from .base import TrainerMixinBase
+
+if TYPE_CHECKING:
```
luyuzhe111: nit: do we need this TYPE_CEHECKING thing?

hellodanylo: nope, removed

luyuzhe111: looks like there are quite a few instances remaining here and there. could you clean up those as well?

src/agentcore_rl_toolkit/backends/verl/examples/swe_agent/main.py
```diff
@@ -0,0 +1,41 @@
...
+)
+
+
+def main(config):
+    auto_set_device(config)
+    config.transfer_queue.enable = True
+
+    config = migrate_legacy_reward_impl(config)
+
+    from agentcore_rl_toolkit.aws_tools.ec2_tools import get_current_instance_type
```
luyuzhe111: Is there a reason for keeping this import inside main()?

src/agentcore_rl_toolkit/backends/verl/examples/swe_agent/main.py
```diff
@@ -0,0 +1,41 @@
...
+
+    with open_dict(config):
+        config.ec2_instance_type = get_current_instance_type()
+        config.trainer.experiment_start_at = dt.datetime.now().isoformat()
+        config.data.train_batch_size = (
+            config.trainer.v1.separate_async.parameter_sync_step * config.actor_rollout_ref.actor.ppo_mini_batch_size
+        )
+
+    run_ppo(config, task_runner_class=TaskRunnerWithRolloutSessionResources)
```
luyuzhe111: should we preserve the validation performed by verl's stock main_ppo before calling run_ppo?

src/agentcore_rl_toolkit/backends/verl/examples/swe_agent/main.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+if __name__ == "__main__":
+    import sys
+
+    config_name = sys.argv.pop()
```
luyuzhe111: can we use Hydra's standard --config-name handling instead of removing the final argument from sys.argv?

src/agentcore_rl_toolkit/backends/verl/examples/swe_agent/README.md
```diff
@@ -0,0 +1,206 @@
...
+## Install
+
+Megatron-Bridge requires Python 3.12. From the repository root:
+
+```bash
+uv sync --python 3.12 --extra verl --extra rollout --group verl-megatron
+```
+
+`--extra rollout` is what the other verl recipes do not need: this loop drives
+`agentcore_rl_toolkit.rollout_session` and `.aws_tools` (aioboto3, backoff), which the
```
luyuzhe111, line 33: regarding aioboto3: maybe we should start using this in other parts of the codebase for async operations in the future too?

also curious, looks like https://github.com/aio-libs/aiobotocore is more regularly maintained than https://github.com/terricain/aioboto3? should we use aiobotocore instead? this is non-blocking

src/agentcore_rl_toolkit/backends/verl/examples/swe_agent/config/main.yaml
```diff
@@ -0,0 +1,33 @@
+hydra:
```
luyuzhe111, line 1: I like moving the recipe away from a large shell command, but are these intended to be independently selectable Hydra config groups? They are currently flat config fragments rather than grouped options such as model: qwen3_30b_p6 or cluster: trainer_rollout_n2, so switching profiles still requires editing the defaults list. Could we either organize the independently variable dimensions as real config groups, or consolidate fragments that are always selected together?

src/agentcore_rl_toolkit/backends/verl/examples/swe_agent/config/qwen3_30b_p5.yaml
```diff
@@ -0,0 +1,35 @@
+actor_rollout_ref:
+  model:
+    path: /opt/dlami/nvme/models/Qwen3-Coder-30B-A3B-Instruct
```
luyuzhe111: maybe use `path: ${oc.env:MODEL_PATH}`? similar for the p6 config

src/agentcore_rl_toolkit/backends/experimental/verl/rollout_session_agent_loop.py
```diff
@@ -0,0 +1,369 @@
...
+                        logger.error(f"Failed rollout {self.session_id} in container: {exception}")
+                        await self.meta.set("aborted", True)
+                except Exception as e:
+                    logger.error(f"Failed rollout {self.session_id} in trainer: {e}", exc_info=e)
+                    exception = exception_to_string(e)
+                    await self.meta.set("aborted", True)
+
+                if self.meta["aborted"]:
+                    agent_loop_outputs = [self.make_failed_loop_output()]
+                await self.save_to_s3(task, agent_loop_outputs, exception, rollout_dump_response)
```
luyuzhe111, line 174: This persistence call is outside the exception handling that converts rollout failures into a placeholder AgentLoopOutput. If the S3 upload or the subsequent DynamoDB update fails after a successful rollout, run() raises without returning any sample, which breaks the guarantee that the GRPO group remains intact. Could observability persistence be made best-effort so that storage failures are recorded but do not discard the training sample?

luyuzhe111 (changes requested):

src/agentcore_rl_toolkit/backends/experimental/verl/trainer_mixins/advantage_metrics.py
hellodanylo: removed another one
in these two files it's actually needed:
src/agentcore_rl_toolkit/concurrency/ray_adapters.py (to avoid ray import during testing)
src/agentcore_rl_toolkit/backends/experimental/verl/trainer_mixins/base.py (to allow mixin user to pick any trainer base)

src/agentcore_rl_toolkit/backends/verl/examples/swe_agent/main.py
hellodanylo: moved to the top

src/agentcore_rl_toolkit/backends/verl/examples/swe_agent/main.py
hellodanylo: good catch!

src/agentcore_rl_toolkit/backends/verl/examples/swe_agent/README.md
hellodanylo: aiobotocore is the underlying library for aioboto3
similarly to how, botocore is used by boto3
the difference is in the level of abstraction
(aio)botocore implements basic service operations, (aio)boto3 adds configuration/resources

> maybe we should start using this in other parts of the codebase for async operations in the future too?

yes, this is definitely preferred over dispatching boto3 in threads

src/agentcore_rl_toolkit/backends/verl/examples/swe_agent/config/main.yaml
hellodanylo: This might be a little subjective, but this setup has been working pretty well for me.
Most of the time I just edit the list of defaults to get a particular configuration that I want.
Occasionally, I also have to add an override in the main config.
So modifying the main.yaml is definitely part of the user workflow.

> are these intended to be independently selectable Hydra config groups

Yes, some of them are mutually exclusive, but the idea is that the user can assemble their own set, including by creating new config variants. For example, if they use p4, they might have another config for p4-specific model settings.

> Could we either organize the independently variable dimensions as real config groups, or consolidate fragments that are always selected together?

I don't quite understand what you mean by this. An example?

src/agentcore_rl_toolkit/backends/verl/examples/swe_agent/config/qwen3_30b_p5.yaml
hellodanylo: good idea.

src/agentcore_rl_toolkit/backends/experimental/verl/rollout_session_agent_loop.py
hellodanylo: Hmm, this is a little debatable. I would rather drop the trajectories than train on tokens that did not get saved for future analysis. Practically speaking, I have not yet seen either S3 or DynamoDB ever fail here. Both have a lot of 9s of availability.

src/agentcore_rl_toolkit/backends/verl/examples/swe_agent/main.py
hellodanylo: awesome, thanks for pointing this out!

src/agentcore_rl_toolkit/backends/verl/examples/swe_agent/config/main.yaml
luyuzhe111: makes sense. I meant something like:

config/
├── model/
│   ├── qwen3_30b_p5.yaml
│   └── qwen3_30b_p6.yaml
├── cluster/
│   ├── rollout_n1.yaml
│   └── rollout_n4.yaml
└── trainer/
    ├── sync.yaml
    └── separate_async.yaml

so these config files are more clearly organized, and ideally easier to navigate for the users too?

literally just learned this from codex yesterday : ) but looks like a good fit here?

src/agentcore_rl_toolkit/backends/experimental/verl/rollout_session_agent_loop.py
luyuzhe111: i see. i would think observability data persistence failure should not fail a rollout. but if it's so rare we are probably good either way!

luyuzhe111 (approved):

src/agentcore_rl_toolkit/backends/verl/examples/swe_agent/config/main.yaml
hellodanylo: I like this! I will do this in a follow-up when I do another verl run.

PR #131, opened 2026-09-09
fix(verl): stabilize optimizer steps for variable-row rollouts

luyuzhe111:
## Summary

This PR adds an `agentcore_sync` verl v1 trainer for rollouts that may produce a variable number of training rows. The trainer keeps the configured actor optimizer-step schedule stable, trains every emitted row, and pads only to the minimum required by actor data parallelism and the configured number of mini-batches.

The implementation uses verl's public trainer registry and `VERL_USE_EXTERNAL_MODULES`; it does not patch verl or replace its workers, engines, TransferQueue, advantage computation, or loss implementation. See the `roadmaps/verl_variable_trajectory_batching.md` for the scheduling model and scope decisions.

## Why

verl v1 currently sends a fixed `mini_batch_size` to the actor worker. When one rollout expands into multiple rows through trajectory forks, context compaction, or sub-agents, the expanded row count can silently increase the number of optimizer steps and require excessive padding.

`agentcore_sync` instead sends a fixed `num_mini_batch`, so optimizer steps remain `data.train_batch_size / actor.ppo_mini_batch_size * ppo_epochs` regardless of row expansion. Padding is reduced to a multiple of `actor_dp_size * num_mini_batches`.

## Changes

- Add and register the actor-only synchronous `agentcore_sync` trainer.
- Preserve verl's existing whole-batch DP token balancing while using fixed mini-batch counts.
- Add batching observability and `training/rollout_failure/missing_sessions`.
- Update the FSDP and Megatron + LoRA recipes to register the trainer through `VERL_USE_EXTERNAL_MODULES`.
- Document the supported configuration and add focused trainer and real-worker regression tests.

## Validation

- `16 passed`: `tests/backends/verl/test_trainer_batching.py` and `tests/backends/verl/test_training_worker_batching.py`.
- Pre-commit passed.
- Completed a 116-step Qwen3-4B GSM8K FSDP run and a 101-step Qwen3-Coder-30B-A3B MigrationBench Megatron + LoRA run.

The GSM8K run improved final validation reward from `0.8605` to `0.9166` (+5.61 percentage points) and avoided the late-training regression visible in the reference run. The MigrationBench best validation reward improved from `0.74916` to `0.78763`, a +3.85 percentage-point / 5.13% relative improvement.

These comparisons provide end-to-end evidence that stable optimizer scheduling improves training behavior under row expansion; in the following screenshots, the green lines have the fix in this PR. 

### GSM8K
Screenshot 2026-09-09 at 3 46 40 PM

### MigrationBench
Screenshot 2026-09-09 at 3 45 26 PM

## Supported scope

`agentcore_sync` currently supports verl v1 actor-only synchronous training with distillation disabled, `parameter_sync_step=1`, and `actor_rollout_ref.actor.loss_agg_mode=seq-mean-token-sum`. Unsupported configurations fail at startup.


src/agentcore_rl_toolkit/backends/verl/trainer.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def _validate_config(config: DictConfig) -> None:
+    if not config.trainer.use_v1:
+        raise ValueError("agentcore_sync requires trainer.use_v1=true")
+    if need_critic(config):
+        raise ValueError("agentcore_sync supports only actor-only training")
+    if is_distillation_enabled(config.get("distillation")):
+        raise ValueError("agentcore_sync does not support distillation")
+    if config.actor_rollout_ref.actor.loss_agg_mode != "seq-mean-token-sum":
```
hellodanylo, line 31: why only this one?

src/agentcore_rl_toolkit/backends/verl/trainer.py
```diff
@@ -0,0 +1,108 @@
...
+        expected_sessions = self.config.data.train_batch_size * self.config.actor_rollout_ref.rollout.n
+        metrics.update(
+            {
+                "batching/real_rows": total_rows - padding_rows,
+                "batching/total_rows": total_rows,
+                "batching/padding_rows": padding_rows,
+                "batching/num_mini_batches": self._num_actor_mini_batches,
+                "batching/required_multiple": self._required_batch_multiple,
+                "batching/configured_optimizer_steps": (
+                    self._num_actor_mini_batches * self.config.actor_rollout_ref.actor.ppo_epochs
```
hellodanylo: the computed config values could also be placed in the params, instead of duplicating in metrics?

hellodanylo (approved):
LGTM

src/agentcore_rl_toolkit/backends/verl/trainer.py
luyuzhe111: ah good question. essentially variable-row expansion changes both the row count and token count. We need every emitted row to add its token losses while retaining the nominal mini-batch denominator; seq-mean-token-sum gives exactly sum(emitted token losses) / nominal_global_batch_size. Token-mean modes instead renormalize by the expanded token count or by each row’s length, changing the objective. With multiple mini-batches, rows from one session may still cross optimizer steps.

luyuzhe111: I've documented both points in the backend README and the variable-row batching design: why seq-mean-token-sum is required, and that with M > 1 rows from one rollout may cross optimizer steps. thanks for the catch!

src/agentcore_rl_toolkit/backends/verl/trainer.py
luyuzhe111: ugh agreed thanks for the catch. I’ll remove these static values.

PR #133, opened 2026-09-10
fix(gateway): add reasoning content to history match dict, enable setting history_mode and linear_on_nonlinear in AgentCoreAgentLoop

lyzustc:
This PR fixes two issues:
- In verl's `AgentCoreAgentLoop`, enabling setting `history_mode` and `linear_on_nonlinear` from init arguments so user can set them from yaml file.
- Add reasoning content to history match dict. A thinking model (e.g. Qwen 3.6-27B) generates reasoning content at one turn, and agent harness may modify reasoning content when constructing next turn (such as remove reasoning content or summarizing reasoning content). This history diverge should be detected at dict match step. If agent harness echoes the same reasoning content to next turn, dict match will pass.


hellodanylo (approved):
LGTM

PR #139, opened 2026-09-12
feat(verl): add gsm8k fsdp lora training example

luyuzhe111:
Add a full-training GSM8K recipe for a single P4d node (8× A100 40GB). Adapt the existing FFT recipe with LoRA rank 16, alpha 32, learning rate 2e-5, and safetensors base-weight loading to avoid the initial weight-sync OOM. All other training settings stay the same.

Validated for one epoch (116 steps), reaching a final GSM8K validation reward of 0.9363: green line is full fine-tuning, red line is lora. 

Screenshot 2026-09-11 at 11 54 19 PM


hellodanylo (approved):
LGTM

PR #140, opened 2026-09-13
feat(verl): expand variable trajectory batching to the async trainer modes

hellodanylo:
# feat(verl): expand variable trajectory batching to the async trainer modes

## Summary

Variable-row batching shipped as a single `PPOTrainerSync` subclass (`agentcore_sync`), so a recipe that needs one of verl's async v1 backends had to choose between a stable optimizer schedule and asynchronous rollout. This PR moves the batching overrides into `VariableRowBatchingMixin` and layers it onto all three verl v1 trainers, registered as `agentcore_sync`, `agentcore_colocate_async`, and `agentcore_separate_async`.

The mechanism becomes trainer-mode agnostic by deriving the actor partition count per `sample -> update` trigger rather than per training step: `M = data.train_batch_size / parameter_sync_step / actor.ppo_mini_batch_size`. That reduces to the previous `train_batch_size / ppo_mini_batch_size` for the colocated modes (both held to `parameter_sync_step=1`) and to `1` for separate-async, which verl already requires to satisfy `train_batch_size == parameter_sync_step * ppo_mini_batch_size`.

## Changes

- `trainer_mixins/variable_row_batching.py` (new): `VariableRowBatchingMixin` carrying the config validation, `_get_required_batch_multiple`, `_update_actor`, and the `batching/*` metrics that previously lived in `trainer.py`.
- `trainer.py` becomes the one definition file: three `register_trainer` names, each layering `AgentLoopMetricsMixin`, `AdvantageZeroMetricsMixin`, and `VariableRowBatchingMixin` onto the corresponding verl trainer. `_AgentCoreTrainerBase` writes verl's own mode string back into `trainer.v1.trainer_mode` before the base trainer initializes, since `PPOTrainer` compares it literally to select the replay buffer, refill semantics, async prompt persistence, TransferQueue checkpointing, and the `trainer.v1.<mode>` node it reads `parameter_sync_step` from.
- The metric mixins move from `backends/experimental/verl/trainer_mixins/` to `backends/verl/trainer_mixins/`, and the swe_agent recipe's `art_*` trainer registrations are deleted — the `agentcore_*` names now carry the same mixins, so recipes no longer need their own trainer module or config-node aliases.
- Docs: the design document and `backends/verl/README.md` are updated for the mixin layout, the per-trigger derivation, and the observability the metric mixins add.

## Behavior notes

- Every `agentcore_*` mode requires v1 actor-only training, distillation disabled, and `loss_agg_mode=seq-mean-token-sum`. All modes but separate-async additionally require `parameter_sync_step=1`: the sync and colocate-async trainers sleep their rollout engines in `on_sample_end` for the whole training pass, so a second trigger would block forever on generation that can no longer happen.
- Layering the mixin onto the previously unvalidated async modes narrows what they accept: a critic, distillation, or another `loss_agg_mode` now fails at startup instead of training under a denominator the fixed partition count does not preserve.
- `agentcore_sync` behavior is unchanged.

## Testing

`tests/backends/verl/test_trainer_batching.py` is parameterized over all three modes: registration in a fresh process, registry-alias normalization, the partition count (per trigger, including a separate-async case with `parameter_sync_step=2`), `_update_actor` metadata and `batching/*` metrics, mixin layering, and the fail-fast config rejections.


src/agentcore_rl_toolkit/backends/verl/trainer_mixins/variable_row_batching.py
```diff
@@ -0,0 +1,144 @@
...
+        actual_sessions = {
+            tuple(key.rsplit("_", 2)[:2])
+            for key, tag in zip(batch.keys, batch.tags, strict=True)
+            if not tag.get("is_padding", False)
+        }
+        # One trigger consumes train_batch_size / parameter_sync_step source prompts.
+        expected_sessions = (
+            self.config.data.train_batch_size // self.parameter_sync_step * self.config.actor_rollout_ref.rollout.n
+        )
+        metrics.update(
```
luyuzhe111, line 141: with `parameter_sync_step > 1`, verl’s [aggregation rules](https://github.com/verl-project/verl/blob/v0.9.0/verl/trainer/ppo/v1/utils.py#L84-L102) sum `total_rows`, but take a weighted average of `real_rows`, `padding_rows`, and `missing_sessions`. this wasn’t visible with sync’s single trigger, but produces inconsistent per-step counts in separate-async mode.

suggested renames:

- `batching/real_rows` → `batching/total_real_rows`
- `batching/padding_rows` → `batching/total_padding_rows`
- `training/rollout_failure/missing_sessions` → `training/rollout_failure/total_missing_sessions`

`batching/total_rows` can stay as-is. These names would make all four counters aggregate by summation.

didn't realize there were different agg modes when i wrote those names originally for the sync trainer : ))))

tests/backends/verl/test_trainer_batching.py
```diff
@@ -22,15 +39,25 @@ def _make_config(*overrides, config_name="ppo_trainer"):
         return compose(config_name=config_name, overrides=[*base_overrides, *overrides])
 
 
+def _trainer(mode, *overrides):
+    """Build the trainer registered under `mode`, with that mode's mandatory config."""
+    _, mode_overrides = TRAINER_MODES[mode]
+    from verl.trainer.ppo.v1 import get_trainer_cls
```
luyuzhe111: Nit: could we move `get_trainer_cls` to the module-level imports?

luyuzhe111 (changes requested):

src/agentcore_rl_toolkit/backends/verl/trainer_mixins/variable_row_batching.py
hellodanylo: great catch!

luyuzhe111 (approved):

PR #141, opened 2026-09-13
fix(gateway): avoid blocking the event loop during tokenization

luyuzhe111:
## Problem and fix

The rollout gateway tokenizes prompts synchronously on its HTTP event loop. While one prompt is being encoded, the loop cannot serve other requests. Long prompts and concurrent agent rollouts amplify this delay.

Use the tokenizer's native `async_encode` and await rendering in gateway callers, preserving HF template and encoding options. This lets the event loop serve other requests while encoding runs. Check session closure after rendering to prevent generation from starting after the session has ended.

## Validation

- **Responsiveness:** eight concurrent 128k-token prompts with the Qwen3-Coder tokenizer, repeated twice. Maximum event-loop scheduling delay fell from 1,205–1,210 ms to 12.7–26.1 ms, with identical token IDs.
- **Token correctness:** exact parity with synchronous rendering across 16 cases using Qwen3-4B and Qwen3-Coder-30B tokenizers.
- **Tests:** 85 gateway tests and 72 verl tests passed. Renderer tests also passed with Transformers 5.0.0.
- **GSM8K e2e performance regression:** completed 10 training steps with Qwen3-4B FSDP LoRA on eight A100 GPUs, tree mode and one AgentLoop worker. Step-10 validation: 87.57% versus 86.42% for the synchronous baseline.


hellodanylo (approved):
LGTM but sync/async code is a mine field for concurrency bugs.

PR #142, opened 2026-09-14
# feat(verl): run `AgentCoreRLApp` agents through the rollout session protocol

hellodanylo:
# feat(verl): run `AgentCoreRLApp` agents through the rollout session protocol

## Summary

This repo has two verl v1 agent loops doing the same job — run one agent rollout somewhere else, capture its tokens through the process-local rollout gateway, hand verl `list[AgentLoopOutput]` — and each has half of what a training run needs. `AgentCoreAgentLoop` (`backends/verl/agent_loop.py`) hardwires its transport to `RolloutClient` + S3 result polling but gets the verl-facing arithmetic right. `RolloutSessionAgentLoop` (`backends/experimental/verl/`) has the pluggable `RolloutSession` protocol, cluster-wide concurrency bounds, timeouts and guaranteed teardown, but reaches only agents that speak the four-POST `wire` protocol, and several of its verl-facing contracts are wrong.

This PR closes both gaps without consolidating the files. `AgentCoreS3Session` reproduces `AgentCoreAgentLoop`'s transport behind the `RolloutSession` protocol — one fire-and-forget `invoke_async`, the session id doubling as the ACR `runtimeSessionId` and the gateway capture key, the result polled out of S3 — so an `AgentCoreRLApp` agent already deployed for batch evaluation trains with no container-side change at all: same image, same payload, same result contract. And `RolloutSessionAgentLoop` absorbs `AgentCoreAgentLoop`'s contracts (token budgets, chat-template kwargs, prompt overflow, staleness tags, `reward_extra_info`), so choosing the pluggable loop no longer costs correctness.

The third piece is what a *failed* rollout contributes to training, and the loop's answer is that the captured tokens decide, not the exception. Any rollout the gateway saw at least one turn from trains — with the container's reward when the rollout finished cleanly, and with a patched `0.0` when either side reported a failure. Only a rollout with **no** trainable tokens is a drop, and what verl is told about a drop is a config field (`on_rollout_failure: raise | empty`, default `raise`) so runs that differ only in that setting are comparable. Every rollout, trained or dropped, is classified into three independent abort flags in its session record and its S3 dump, so a run's failure mix is queryable rather than inferred from log lines.

The fourth is the AgentCore data plane itself. Rollout sessions are constructed one per trajectory, and a `bedrock-agentcore` client per session is a connection pool per rollout; both ACR-backed sessions now ride one process-wide client instead. That client is also what makes the new S3 backend's microVM warm-up affordable, and the warm-up is where the platform's readiness race surfaces — retried here, because botocore will not.

### The `agentcore_s3` session backend

`rollout_session/agentcore_s3_session.py` (new) implements `RolloutSession` over `RolloutClient`.

`setup` starts the ACR session ahead of the rollout with a trivial warm-up command and records the runtime and result bucket into the session record. The invoke would create the session itself, but then the microVM cold start would be charged to `agent_run_timeout`; starting it in `setup` puts it under `container_setup_timeout`, where the other backends' provisioning lives. Preparing the *task* inside the container still has no channel of its own here — whatever the handler does before the agent runs is `agent_run_timeout`'s.

`run` invokes once and awaits the `RolloutFuture`. `shutdown` stops the ACR session and is idempotent, so it is safe whether the rollout finished, timed out, or never started; two paths reach the same session. When `run` got a future, the stop rides `RolloutFuture.cancel_async`, itself a no-op once the result fetch stopped the session. When no future exists — a `setup` that raised, an invoke that raised after ACR accepted it — the stop is made directly against the runtime rather than leaving a warm microVM to ACR's idle reaper. A `_started` flag, set *before* the warm-up call rather than after, is what shutdown acts on in that case: a warm-up that fails is precisely the case where the microVM usually does exist.

Around that core the module carries the conventions the `@rollout_entrypoint` result dict actually uses, lifted from `backends/verl/agent_loop.py`: `status_error` turns a non-200 `status_code` into the dump's `exception`, `extract_agent_reward` reads the `{"rewards": ...}` convention (scalar, or last element of a list), and `numeric_metrics` keeps the float-valued entries of the agent's `metrics` for the dump while everything else survives in `task_output`.

`result_location(rollout_output_s3, experiment_name)` derives `RolloutClient`'s `s3_bucket` / `exp_id` pair from the config's canonical names. The agent SDK writes `{exp_id}/{input_id}/{session_id}.json` in one bucket and knows nothing of prefixes, so a prefix in the location is folded into `exp_id` rather than silently dropped: `s3://bucket/runs` with experiment `exp-1` puts a rollout at `s3://bucket/runs/exp-1/<group_id>/<session>.json`, alongside the loop's own dumps.

`get_or_create_rollout_client` caches one client per resolved config, because a session is constructed per trajectory while the client owns the boto3 clients and their connection pools; `reset_client_cache` exists for tests. Two other guards are local to this module: the client's own TPS limiter is set effectively infinite (`UNTHROTTLED_TPS`), since ACR throttling belongs to `bounds.session_rate_limiter` — a cluster-wide named actor, and therefore the only place a per-ARN cap can actually be enforced — and a session id shorter than ACR's 33-character `runtimeSessionId` floor is rejected at construction rather than surfacing as a botocore `ValidationException` one rollout later.

`rollout_session/factory.py` gains the `agentcore_s3` branch and its keys (`rollout_output_s3`, `experiment_name`, `max_pool_connections`, defaulting to 100 rather than `RolloutClient`'s 10); `agentcore_runtime_arn` is shared with the HTTP backend, since it is the one name for an ACR runtime. There is deliberately no `tps_limit` knob.

### `agentcore` → `agentcore_http`

The existing session is renamed (`agentcore_session.py` → `agentcore_http_session.py`, `AgentCoreSession` → `AgentCoreHttpSession`, backend name `agentcore` → `agentcore_http`) now that two backends reach ACR by different routes. This touches the factory, the EC2-monitor guard in `rollout_session_resources.py`, the `swe_agent` recipe config and README, and the `openhands_swe_agent` example and its tests. **Existing configs naming `backend: agentcore` must be updated.**

### One AgentCore data-plane client per process

`aws_tools/agentcore_tools.py` grows `shared_agentcore_client(region)`: the process-wide `bedrock-agentcore` client, opened on first use, closed by no caller. Both ACR sessions now make every call — start, setup, each status poll, the dump, the stop — over it, and neither owns its lifetime, so every phase works whether or not the session was entered and `shutdown` stays safe on a session that never ran. `AgentCoreHttpSession` loses its `AsyncExitStack` entirely as a result, along with the exception-juggling its teardown needed to avoid erasing the rollout's own error; implicit chaining now does that, and `describe_with_root_cause` digs the cause out for the log line.

The cache is keyed by `(running loop, region)`, not region alone: a client's aiohttp connector belongs to the loop that created it, and the agent-loop worker, the driver and each test have their own. Creation is guarded by one lock per loop, or the first burst of rollouts would each open a client and all but one would be dropped unclosed. `close_shared_agentcore_clients()` ends the current loop's clients — for tests, and for a worker that ends its loop while the process continues.

`SESSION_CLIENT_CONFIG` now also sets `max_pool_connections=512`. That pool sizes a whole worker's ACR concurrency rather than one rollout's, and a rollout holds at most one connection at a time (its phases are sequential), so it covers the example configs' `container_concurrency: 256` with room for the teardown calls beside it; botocore's default of 10 would queue session starts. Connections open lazily, so unused headroom costs nothing.

### Two stops, and the readiness retry

`stop_agentcore_session` splits by what owns the session slot. `stop_agentcore_instance_session` is the capacity-provider delete, for sessions placed on a pooled instance where failing to delete leaks one instance until its idle timeout; an already-deleted session is not an error, anything else raises. `stop_agentcore_microvm_session` is the plain runtime stop for a session owned by nothing but the rollout that started it, and it never raises: a stop happens on the way out of a rollout, often carrying the error that ended it, and a teardown failure that replaced that error would cost the diagnosis to save a session ACR's idle reaper collects anyway.

`start_agentcore_session`'s warm-up is now retried through one specific failure. ACR creates the microVM and the container does boot, but a command dispatched before the app is serving comes back as `RuntimeClientError: Received error (500) from runtime`. The retry has to live in this module rather than in the client's config: `RuntimeClientError` is modelled `httpStatusCode: 424, senderFault: true`, so botocore reads it as the caller's fault and `SESSION_CLIENT_CONFIG`'s `max_attempts` never fires on it. Three tries, 2s apart, no jitter — the interval is the measured recovery time, not a herd-spreading guess — and both tunables are read through callables so tests can turn the interval off without reaching inside the decorator. `is_runtime_readiness_error` matches by error code both ways it arrives: as the modelled exception on the initial response (what every observed failure was) and wrapped in an `EventStreamError` mid-stream. Nothing else is retried: every other way a start fails is either already retried by botocore (throttling, 5xx *service* errors) or permanent for this rollout, where more tries would only multiply the burst.

### `RolloutContractError`, and `group_id` as the required task field

`rollout_session/errors.py` (new) introduces the one distinction a trainer needs: a *failed rollout* (container never came up, handler raised, run timed out) is expected at some rate, so what to report for it is a policy choice, while a *contract violation* (a dataset with no `payload` column, a row with no group id, a reward function returning a string) recurs on every rollout of the run. It subclasses `ValueError`, so existing `except ValueError` callers are unaffected. Nothing treats the marker type specially inside the loop yet — it is raised, logged with its traceback and recorded in the S3 dump like any other rollout exception, which is enough to diagnose a run that fails on every row.

The required per-task identifier is the rollout **group** id, not `task_id`. `AgentCoreS3Session` keys its result objects by it (`{exp_id}/{group_id}/{session_id}.json`), so a group's rollouts land in one prefix, and `_require_group_id` refuses to fall back: the row index keys results by position, which shuffles between runs, and the session id is unique per rollout, so either one scatters a group's results instead of grouping them. The name is the harness's contract rather than any one trainer's — the verl loop maps its own `uid` onto it, and an evaluator with no notion of groups can pass the task's own id, one rollout per group. `task_id` is optional: recorded as a coordinate in the session record when present, read by no session.

### `RolloutSessionAgentLoop`: the merged verl contracts

Startup now fails fast instead of misbehaving later: `trainer.use_v1` is required (the loop returns a list, which only the v1 TransferQueue path consumes), an `on_rollout_failure` typo is rejected at construction rather than hours into a run, and `_validate_token_budgets` rejects budgets a trajectory could not be stored in — an explicit positive `rollout.max_model_len`, `prompt_length` / `response_length` within it, and a `rollout_gateway_sampling_params.max_new_tokens` that is a positive int within it. Lengths now come from `actor_rollout_ref.rollout.*` rather than `data.max_*`, because those are the widths of the regions a trajectory is actually stored in. The constructor takes verl's arguments explicitly rather than `*args`, and the trainer's `apply_chat_template_kwargs` (e.g. `enable_thinking`) now reach the gateway renderer, so it tokenizes turns exactly as training will — a functional regression against `AgentCoreAgentLoop` that had gone unnoticed.

Record-to-output conversion changes on three points. Every trajectory-tree leaf is emitted, not just the most-trained one, with the primary record placed **last** because verl scores and broadcasts from a session's last output. An initial prompt that exceeds `prompt_length` now has its overflow placed at the head of the response region, masked and with zero logprob, rather than being truncated away — the token order stays complete and attended to while carrying no policy loss. Filler positions use the tokenizer's EOS instead of literal token `0`, and the stand-in response token for a prompt that fills the whole budget is masked (`response_mask=[0]`, previously `[1]`, which trained a loss on a synthetic token). `num_turns` gains the initial user turn that verl counts and the gateway never sees.

Staleness and metrics are made honest. `min/max_global_steps` fall back to the dispatch step when the engine reports none (all turns served before a restart) rather than raising, and a weight version of `0` is no longer mistaken for absent; the dispatch step itself defaults to `0` rather than the old `-1` sentinel, which v1's staleness metrics would have consumed as a real step. Session-level totals (`llm_generated_length`, `context_length`) are computed once for the session instead of once per leaf. `reward_extra_info` now carries `reward_score` under a canonical key (DAPO group filtering reads it, and the v1 replay buffer raises when a finished trajectory lacks the configured key), plus `num_trace_records`, plus exactly the agent metrics *declared* in the new `reward_extra_info_defaults` config field — declared rather than discovered, because verl reduces this dict across the batch and a key only some rollouts report is silently averaged over the wrong denominator. The key set is identical whether the rollout succeeded or failed.

### Abort classification: what trains, and what verl is told

Every rollout ends with three independent flags in its session record and its S3 dump, plus the score they imply:

* `aborted_by_trainer` — an exception on this side of the wire (setup failed, the bounded run timed out, the session raised) prevented getting the agent's terminal output.
* `aborted_by_agent` — a dump came back that does not describe a finished rollout: the container reported an exception, or reported no reward, or no task output. `RolloutDumpResponse.failure_reason` is the single gate.
* `aborted_by_gateway` — nothing trainable was captured: no turn reached the gateway, or every record came back with no token ids.

The first two are mutually exclusive (either the container answered or it did not); the third is independent of both. `aborted` is their disjunction, and `reward_patched` records whether the trained score is the container's or a substituted zero.

What trains follows from the flags rather than from where the exception was raised. **Any rollout with captured tokens is trained**: with `float(dump.reward)` when nothing aborted, and with a patched `0.0` when either the agent or the trainer side aborted. This is the dominant abort — a run that timed out with turns already captured — and the loop deliberately does not let its own exception decide what trains when the tokens are sitting there; a partial trajectory at a defaulted zero is a real observation, and the exception is recorded beside it. **A rollout with no captured tokens is a drop**: nothing trained, so its score is recorded as absent rather than as a zero that would read as a real one, and `on_rollout_failure` decides what verl hears.

The rest of the failure handling is arranged around that. The S3 dump is written *first* in every mode, so a drop is on record whatever happens next, and the whole session record goes into it — spans, timeout flags and all — because a discarded rollout contributes nothing to the step's `agent_loop/*` metrics and the dump is the entire account of it. The gateway session is always drained: `create_session` sits outside the `try`, so a setup failure does not `finish_session` an id that was never created, and turn counting and `pop_extra_fields` happen in the `finally` so they survive every exit.

### `on_rollout_failure`

The field decides what a *drop* — a rollout with no trainable tokens — reports to verl. Both settings write the S3 dump and the DynamoDB session record identically; they differ only in what reaches the trainer.

`raise` (the default) raises out of the loop, and `_run_prompt` marks the prompt group `failure`. The sync buffer leaves such a group sampleable, so the healthy siblings still train; `ReplayBufferAsync` treats `failure` as terminal and evicts *and refills* the whole group, discarding those siblings and re-running the prompt.

`empty` returns no rows: `_agent_loop_postprocess` logs the empty output and writes nothing to the TransferQueue. The group stays `finished` with fewer than `n` rows and every trainer mode trains the survivors — but no `extra_fields` are written either, so every `agent_loop/*` number about the drop vanishes from the step, and a group whose rollouts *all* drop raises anyway (from DAPO's group filter, or from the sync buffer's "no materializable trajectories" guard).

Which one trains better is an open question; the field exists to gather comparable runs, not because one is known to win. A third option — a masked zero-reward row shaped like verl's own padding, which would keep the metrics and the group's full width — is deliberately not offered: it puts a synthetic zero inside the GRPO group, so `[1, 1, 1]` plus a drop trains as `[1, 1, 1, 0]` and every sibling's advantage shifts because a container died. Isolating such a row in a group of its own (an earlier trainer-side mixin on this branch) trades that for a group of one, mean 0 / std 1 — the raw reward as its own advantage, with no baseline — which is its own bias. Neither is in this PR.

### CI, docs, recipe config

The verl integration workflow now runs `tests/backends` rather than `tests/backends/verl`, so the experimental backend's tests are covered, and watches `backends/experimental/verl/**` and `rollout_session/**`.

## Behavior notes

`agentcore_s3` buys agent-side compatibility at a cost, all of it accepted deliberately and documented in the module docstring. The setup/run split is thin — `setup` can only start the microVM early, not prepare the task — so cold start is bounded by `container_setup_timeout` while in-container task preparation is charged to `agent_run_timeout`. There is no status channel, so a container that dies without the SDK's error path running (OOM, process death) is noticed only when `agent_run_timeout` expires; a handler that merely raises still comes back promptly through S3 as `{"status_code": 500, ...}`. Completion is polled, so a finished rollout is noticed up to `RolloutFuture`'s 30s `max_interval` late. And reward and metrics are conventions of the result dict rather than typed dump fields.

Two reward-path departures from `AgentCoreAgentLoop`, both intentional: a *successful* result carrying no reward is treated as a failed rollout rather than a 0.0 at the session layer (the loop then decides, by whether tokens exist, between a patched zero and a drop), and a non-numeric reward raises `RolloutContractError` rather than scoring 0.0 — broken reward code is broken on every rollout, and the zeros would flatten every group's advantages.

The loop never propagates a rollout's own exception to verl. A trainer-side failure with tokens trains at zero; without tokens it becomes a `RuntimeError` naming the missing token records under `raise`, or an empty list under `empty`. The original exception is in the log line (with its root cause) and in the dump's `trainer_exception`.

`on_rollout_failure` interacts with the trainer mode, so the two should be chosen together: `raise` is cheap under `agentcore_sync` and expensive under either async mode, which refill the whole group.

ACR throttling for the `agentcore_s3` backend lives entirely in `bounds.session_rate_limiter`. Size it for roughly three calls per rollout: the setup warm-up, the invoke, and the teardown stop, which shares ACR's per-ARN budget but happens outside the bounded region — with the client's adaptive boto3 retries as the backstop.

## Testing

`tests/backends/experimental/verl/` is new (54 tests, with a shared `conftest.py` faking the session, the gateway and S3), organized along the axes the loop now has: the construction contracts (v1, the length budgets, an unknown failure mode, chat-template kwargs precedence); the successful path (end-to-end single trace, multi-turn merge, turn counting before the drain, `uid` becoming the group id and driving bounds priority, sampling defaults, the session token budget, staleness fallback and the zero-weight-version case, multi-leaf emission with the primary last); what the reward becomes when tokens *were* captured (a dump with no reward, no task output, a container-reported failure, a trainer-side exception, a `RolloutContractError` — all trained at a patched zero, with the flags asserted); and what verl is told when they were *not*, parameterized over every way that happens (no turns, a container-reported failure before the first turn, a trainer-side failure before it, and a stale agent image capturing under a static session id) crossed with both `on_rollout_failure` settings. Alongside those: the dump written before the raise and carrying the whole session record, the gateway session drained on every exit including a setup failure, prompt overflow, the inert response for a budget-filling prompt, and `reward_extra_info` / `extra_fields` key-set stability.

`tests/rollout_session/agentcore_s3_session_test.py` adds 53: dump translation, agent-reported failure, the invoke arguments (session id doubling as the capture key, `group_id` as `input_id`, the stripped `openai/` prefix, the verbatim payload, no forwarded sampling params), the setup warm-up and where it records having run, the task contracts (including a task with no `task_id` still running), shutdown along all five paths (finished, idempotent, set-up-only, setup-failed, failing stop not replacing the rollout's error), `result_location`, the disabled client throttle, the client cache, and the factory branch. The timeout case drives the real `run_rollout_with_bounds` with local semaphores rather than a bare `asyncio.timeout`, so teardown-while-unwinding is what is actually exercised.

`tests/aws_tools/agentcore_tools_test.py` is new (20): the shared client's identity per region, a concurrent first-caller burst opening exactly one, per-region and per-loop isolation, the config it carries, closing and reopening; then the warm-up's readiness retry (retried against the same session, mid-stream error events too, giving up after the last try, everything else raising on the first try, a non-zero exit code being a start failure), the best-effort microVM stop, and a check that the pool covers the concurrency the example configs ask for. `agentcore_http_session_test.py` is reworked (20) onto the shared client.

`tests/backends/verl/test_advantage_metrics.py` (3) is carried along from this branch's earlier work: it pins what `critic/advantages/zero_*` counts and that it ignores verl's padding rows.

Locally: `pytest tests/backends tests/rollout_session tests/aws_tools` → 338 passed.


src/agentcore_rl_toolkit/rollout_session/lifecycle.py
```diff
@@ -18,9 +18,29 @@
...
 from agentcore_rl_toolkit.concurrency.priority_semaphore import PrioritySemaphore
 from agentcore_rl_toolkit.concurrency.rate_limiter import RateLimiter
+from agentcore_rl_toolkit.rollout_session.errors import RolloutContractError
 from agentcore_rl_toolkit.rollout_session.wire import RolloutDumpResponse
 
 
+def require_task_id(task: dict) -> str:
+    """The dataset's own id for this task, or a loud error.
+
+    Every row must carry ``task_id``: it is what rollouts are grouped by, what lands in
```
lliquid: this could be a general dataset format check across all training backends and use cases ?

src/agentcore_rl_toolkit/rollout_session/agentcore_s3_session.py
```diff
@@ -0,0 +1,308 @@
...
+    max_pool_connections: int = 100,
+) -> RolloutClient:
+    """The process-wide :class:`RolloutClient` for this config, created on first use.
+
+    Takes the config's canonical names; ``RolloutClient``'s own ``s3_bucket`` / ``exp_id``
+    pair is derived from them by :func:`result_location`.
+    """
+    bucket, exp_id = result_location(rollout_output_s3, experiment_name)
+    key = (agentcore_runtime_arn, bucket, exp_id, max_pool_connections)
+    client = _CLIENTS.get(key)
```
lliquid: is this for rollout different task with different AC arn ? i am wondering if we can have a high level abstraction for it. Or is it more of a utility function.

src/agentcore_rl_toolkit/rollout_session/agentcore_s3_session.py
```diff
@@ -0,0 +1,308 @@
...
+    return bucket, f"{prefix}/{experiment_name}" if prefix else experiment_name
+
+
+def reset_client_cache() -> None:
+    """Forget every cached client. For tests; each client holds boto3 clients."""
+    _CLIENTS.clear()
+
+
+class AgentCoreS3Session(RolloutSession):
+    """A rollout session backed by one ACR invocation of an ``AgentCoreRLApp`` agent."""
```
lliquid: the main logic here is not backend (verl) specific ?

lliquid: ah missread, it is not in the backend folder.

luyuzhe111 (approved):
from codex:

1. **[P1] The default p6 configuration fails the new token-budget validation**

   The default `main.yaml` selects `qwen3_30b_p6`, but this PR only adds `rollout.max_model_len` to the p5 configuration. The composed p6 configuration still has `rollout.max_model_len=None`; its `engine_kwargs.vllm.max_model_len=64000` does not populate the agent loop’s configuration. Consequently, every rollout using the default recipe fails during initialization. Update the p6 configuration as well and add a test that constructs the loop from the default recipe.

2. **[P1] `RolloutContractError` does not terminate asynchronous training**

   Re-raising this exception still allows verl 0.9.0’s `_run_prompt` to catch it and publish `status=failure`. `ReplayBufferAsync` then evicts and replaces the failed group. I reproduced the worker returning normally instead of propagating the exception. If every dataset row lacks `payload`, or the reward implementation consistently returns invalid values, asynchronous training can keep failing and refilling indefinitely. Validate static contracts before training starts and provide a trainer-visible termination path for runtime contract violations.

3. **[P2] Timing out before invoke returns leaks the runtime session**

   `_future` is only assigned after `invoke_async` returns. If ACR accepts the request but cold start or response reading exceeds `agent_run_timeout`, the underlying `asyncio.to_thread` call continues running while `shutdown()` returns immediately because `_future` is still `None`. Using the real `RolloutClient` with a mocked service response, I reproduced a completed invocation with zero `stop_runtime_session` calls. Track the pending invocation and ensure cancellation eventually cleans up the session. The existing timeout test only covers polling after a future has been obtained.

4. **[P2] Waiting for a rollout slot bypasses the intended request-rate limit**

   `run_rollout_with_bounds` consumes the rate-limit permit before waiting for `rollout_semaphore`, but the new backend invokes ACR only after acquiring that semaphore. When container concurrency exceeds rollout concurrency, requests can pass the limiter, accumulate in the queue, and then invoke in a burst as slots become available. This PR also disables the client-side limiter. With a configured limit of 10 TPS, I reproduced invocation intervals of only 0.03–0.06 milliseconds. Apply the shared rate limiter immediately before the actual invocation.

5. **[P2] The default validation dataset is replaced with the training dataset**

   `val_files` now points to exactly the same parquet file as `train_files`, replacing the previous SWE-bench verified validation dataset. The default recipe validates before training and every 10 steps, so those metrics will measure performance on training tasks and no longer provide an independent assessment of generalization. Preserve a separate validation dataset; if evaluation on training tasks is intentional, expose it through an explicit, separate configuration.
 
if you can fix comments that are actually relevant, the PR should be good to go!

hellodanylo:
Fixed issues 2 and 4, as well as handled timeout+tokens as a learnable data point.
Recipe updates will come in the next PR.

PR #143, opened 2026-09-14
perf(gateway): encode linear history deltas from rendered text

luyuzhe111:
## Summary

Linear healing tokenizes the full conversation four times per appended turn: the stored history, its extension, and two probes that identify the assistant's closing delimiter. For long histories, tokenization is much more expensive than template rendering: in a separate synthetic Qwen3-Coder measurement at approximately 100k tokens, rendering took **0.15 ms** and tokenization took **137.03 ms**。

This change retains full-history template rendering to preserve context-dependent formatting, while tokenizing only the closing delimiter and newly appended text. It builds on [#141](https://github.com/awslabs/agentcore-rl-toolkit/pull/141), which separates template rendering from async tokenization.

## Implementation

`HfTemplateRenderer.render_delta()` renders the real stored history and its extension, checks their text prefix, and extracts the new tail by character position. It identifies the assistant closer using the existing two-probe approach in text space. All four template renders still run; only the closer and tail pass through tokenization.

`LinearHealer` appends these token IDs to the actual served prefix, using its existing overlap trimming for closer tokens already present. If delta extraction cannot establish the prefix or closer, it falls back to the existing full-history healer. The shared probe constants and common-suffix helper serve both paths.

Text slicing also handles append boundaries that BPE merges across. For example, Qwen3-Coder encodes `\n` as `[198]` and `\n\n` as `[271]`. With a template that places a newline on each side of the append boundary, full encoding loses the original token prefix. Extracting the new newline in text and encoding it separately preserves the served token IDs. This boundary case is covered by a synthetic-template regression test.

## Validation

- **95 gateway tests passed**, including delta extraction from real history, preservation of generated tokens and training masks across turns, fallback behavior, and BPE boundaries.

Offline benchmark of `LinearHealer.heal()` using synthetic Java/tool text and the Qwen3-Coder-30B-A3B-Instruct tokenizer. Both variants use async tokenization. At **concurrency 1**, values are medians of seven timed iterations after warm-up.

| History tokens | New tool-result tokens | Full-history healer | Delta healer |
| ---: | ---: | ---: | ---: |
| 12,844 | 21 | 51.08 ms | 0.61 ms |
| 50,644 | 21 | 204.84 ms | 0.95 ms |
| 101,044 | 21 | 414.94 ms | 1.48 ms |
| 101,044 | 50,400 | 468.10 ms | 51.86 ms |

At **concurrency 8**, eight independent sessions share one renderer and event loop. Each batch releases eight requests together using the same input size, then waits for all eight to finish. After one warm-up batch, seven timed batches yield 56 request samples per variant and input size. The table reports median request latency from the common release, including event-loop scheduling.

| History tokens | New tool-result tokens | Full-history healer | Delta healer |
| ---: | ---: | ---: | ---: |
| 12,844 | 21 | 101.54 ms | 4.10 ms |
| 50,644 | 21 | 395.92 ms | 5.48 ms |
| 101,044 | 21 | 804.00 ms | 8.98 ms |
| 101,044 | 50,400 | 948.37 ms | 141.15 ms |

Every concurrent result matched the full-history healer and preserved the served token prefix.


lyzustc (approved):

hellodanylo (approved):
LGTM but I wish we had a sample of token data to test the gateway with.

luyuzhe111:
@hellodanylo sounds good. will add it in a separate gateway ci pr!

PR #144, opened 2026-09-14
feat(verl): Support Qwen3.6 model families. Add reward threshold.

Linbo-Liu:
## Summary

Adds a working Megatron + LoRA GRPO recipe for **Qwen3.6-27B** on the MigrationBench agent, plus the four fixes it needed. Qwen3.6-27B is `Qwen3_5ForConditionalGeneration`: a *dense* hybrid model (64 layers of 3× GDN `linear_attention` + 1× gated `full_attention`, standard MLP) that also carries a `vision_config`. Both of those properties break assumptions in recipes written for text-only MoE models, and three of the four fixes below trace directly to the vision-language architecture rather than to anything about the data — the task here is text-only.

## What was broken

| Symptom | Root cause |
|---|---|
| `Triton Error [CUDA]: misaligned address` from the GDN backward, on the first optimizer step | fla 0.5.1's `gated_delta_rule` backward faults when `H == HV`; this model expands q/k to 48 heads via `repeat_interleave`, so all 48 `linear_attention` layers hit it |
| `RuntimeError: The expanded size of the tensor (N) must match the existing size (M)` from `actor_rollout_compute_log_prob` at any `context_parallel_size > 1` | megatron-bridge's `qwen_vl` `preprocess_packed_seqs` slices chunk one of the zigzag-CP split by position in the *padded* sequence while reading a buffer of real tokens only, and clamps chunk two but not chunk one |
| `torch.OutOfMemoryError: Tried to allocate 199.25 GiB` in `update_actor` | the VL path densifies the batch to `[b, max_seqlen]` *before* the THD repacking, so memory scales with `b × max_seqlen` rather than total tokens, defeating remove-padding |
| Every session forked into two training rows, both credited with the same reward, one carrying an all-zero `loss_mask` | default `history_mode="tree"` forks on re-tokenization drift; `num_trace_records` read exactly 2.0 across all rollouts |

Two of these are especially expensive to discover because they surface *after* a full rollout is already paid for: the CP crash and the OOM both land in the training-side passes, so a single bad row or an unlucky micro-batch costs a ~45–90 minute batch.

## Changes

**`flash-linear-attention>=0.5.2`** (`pyproject.toml`, `uv.lock`) — fixes the GDN backward fault. Scope note: only the gated-delta-net path imports fla (`megatron/core/ssm/gated_delta_net.py`, `verl/models/transformers/qwen3_5.py`), so standard-attention models such as Qwen3-Coder-30B-A3B never execute these kernels and are unaffected either way.

**`patches/megatron-bridge-cp-chunk1-clamp.diff`** + an idempotent applier — clamps chunk one the way chunk two already is, which unblocks `context_parallel_size > 1`. Verified byte-identical on normal-length rows, so it is a no-op except on rows too short to split. The short rows are not avoidable from config: verl synthesizes `prompt_len=1 / response_len=1` samples itself in `trainer/ppo/padding_utils.py` for dp divisibility. Nor is any `cp > 1` layout safer than another, since the alignment depends on the `tp × cp` product alone. Shipped as a patch file rather than a hand edit because `uv sync` reinstalls the package and reverts it.

**`reward_thresholds` and `agent_reward`** (`agent_loop.py`) — `MigrationReward` is graded (0.0 build failed / 0.5 built but tests not preserved / 1.0 both), so its mean is not a pass rate: 0.67 is equally 67% full passes or 34% full passes plus 66% half credit. `reward_thresholds` is a `{name: threshold}` map that emits `reward_extra_info[name] = 1.0 if reward >= threshold`, recovering per-stage success rates without the agent reporting them (no container rebuild). `agent_reward` carries the agent's scalar as a plain per-row float, deliberately not keyed `reward` — verl fills that key itself from `rm_scores`, and writing the same key appends to its list rather than replacing it.


patches/megatron-bridge-cp-chunk1-clamp.diff
```diff
@@ -0,0 +1,27 @@
+--- .venv/lib/python3.12/site-packages/megatron/bridge/models/qwen_vl/modelling_qwen3_vl/utils.py.orig-preclamp	2026-09-10 06:54:42.000000000 +0000
```
luyuzhe111, line 1: thanks for figuring this out!

src/agentcore_rl_toolkit/backends/verl/agent_loop.py
```diff
@@ -321,6 +328,10 @@ def _reward_extra_info(self, result: dict[str, Any] | None, n_records: int, *, f
                         info[k] = float(v)
                     except (TypeError, ValueError):
                         continue
+
+        info["agent_reward"] = float(reward)
```
luyuzhe111, line 336: is there any consumer for this metric?

src/agentcore_rl_toolkit/backends/verl/agent_loop.py
```diff
@@ -157,6 +159,9 @@ def __init__(
         self._rei_defaults = {str(k): float(v) for k, v in (reward_extra_info_defaults or {}).items()}
         # verl already derives its `reward` validation metric from rm_scores.
         self._rei_defaults.pop("reward", None)
+        self._reward_thresholds = {str(k): float(v) for k, v in (reward_thresholds or {}).items()}
+        for name in self._reward_thresholds:
+            self._rei_defaults.setdefault(name, 0.0)
```
luyuzhe111: do we need this loop? Threshold metrics are computed and assigned on every output, so they do not need entries in `_rei_defaults`, which handles defaults for agent-provided metrics?

Instead, maybe consider reject overlapping names between `reward_thresholds` and `reward_extra_info_defaults` at initialization. This keeps the two metric sources separate and prevents a declared agent metric from being silently overwritten by the threshold calculation.

src/agentcore_rl_toolkit/backends/verl/examples/migration_agent/agentcore_agent.yaml
```diff
@@ -7,5 +7,11 @@
...
   max_tokens_per_turn: 8192
   max_rollout_time: 1800
-  # Preserve sampled tokens by forking rather than realigning across turn drift.
+
+  history_mode: linear
+  linear_on_nonlinear: reset
   fork_threshold_tokens: 0
+
+  reward_thresholds:
+    pass_at_1: 1.0
```
luyuzhe111: nit: consider renaming `pass_at_1` to `task_success`. This field represents a binary outcome for one rollout (`reward >= 1.0`); pass@1 is the aggregate statistic derived from those outcomes. `task_success` describes the stored value more directly and pairs naturally with `build_success`.

luyuzhe111 (changes requested):

lliquid (approved):

luyuzhe111 (approved):
LGTM. comments are mostly nits and not blocking.

src/agentcore_rl_toolkit/backends/verl/agent_loop.py
Linbo-Liu: This is for logging purpose. `agent_reward` is raw reward from ACR session. I found `rm_score` has been parsed by verl so sometimes it doesn't match the raw reward due to padding issue.

src/agentcore_rl_toolkit/backends/verl/agent_loop.py
Linbo-Liu: Thanks for the comment! Makes sense.

src/agentcore_rl_toolkit/backends/verl/agent_loop.py
luyuzhe111: ohh got it. thanks for the explanation!

PR #150, opened 2026-09-17
fix: forward chat template kwargs in linear history mode

luyuzhe111:
Linear history mode ignored request-level `chat_template_kwargs`, so options such as `enable_thinking` could not override renderer defaults. Forward these kwargs through initial rendering, incremental healing, assistant-closer probes, and full-render fallbacks.

Each session retains its last committed kwargs. If effective kwargs change, compare the same stored history under both configurations; a changed history follows `linear_on_nonlinear` (default: reset), while identical history continues healing. Unchanged effective kwargs skip the extra old-history render.

## Validation

- 73 tests passed across renderer, linear healer, adapter, and gateway integration tests in an environment synced with `uv sync --frozen --extra dev`.
- Regression tests cover request overrides, changes that affect history versus those that do not, skipping the extra render for unchanged effective kwargs, restoring defaults when kwargs are omitted, session isolation, and reset paths using the current kwargs.
- Ruff 0.1.5 lint and format checks, and `git diff --check`, passed.


hellodanylo:
LGTM though it would be great to document which models/harnesses can trigger this path.

lyzustc (approved):

luyuzhe111:
we have decided to not support per-turn chat template kwargs update and use the initial config passed to gateway as the source of truth

PR #152, opened 2026-09-18
fix(gateway): use initialization-only chat template kwargs

luyuzhe111:
Request-level `chat_template_kwargs` could override the gateway renderer's initial configuration. Remove request-level forwarding and per-call renderer overrides so rendering consistently uses the configuration supplied at initialization, preserving the verl configuration path.

Adapters ignore request-level kwargs and warn once per adapter instance when a nonempty value is supplied. Omitted values, `null`, and `{}` remain silent, and requests continue normally.

## Validation

- Gateway test suite: 99 passed.
- Manual OpenAI and Anthropic request checks verified empty-value handling, one warning across repeated requests, independent adapter instances, and successful responses.
- Ruff 0.1.5 lint and formatting checks, plus `git diff --check`, passed.


hellodanylo:
LGTM!

hellodanylo (approved):

